PR #180, opened 2021-11-02
device fix

jasonkhadka:


sjrl (approved):

jasonkhadka:
For now, reverting Unbatchify to have cpu=True by default, and in `if not use_post` `cpu=False`.
We have to create new issue for thinking about better device handling.

Issue #194, opened 2021-11-03
Allow serializer and deserializer to be passed to value.


blythed:
We could then save these functions using the serializer itself.

wuhu:
This could work like this:

```python
def mysave(value, path):
    [... define how to save *value* at *path* ...]
 
def myload(path):
    [... define how to load from *path* ..., should return *value* ...]

[...]

a = value(x, serializer=(mysave, myload))
```

The serializer will call the save-function upon saving and put the load-function in the `transforms.py` then use it to load.
Therefore, we need to get the source of load-function (can be done with `inspect.getsource`).

Problem could be that there is no check that it will actually load. We could (should?) add a check - try loading when saving.

wuhu:
done in #209 

Save function must look like this, takes the path, an integer "i" that can be used to construct a filename and must return a filename:

```
def mysave(val, path, i):
    filename = ...
    return filename
```

The load function takes a path and returns the value
```
def myload(path):
    return ...
```

These can then be used via

```
value(x, (mysave, myload))
```

blythed:
Complex case:

```python
def mysave(val, path):
    return [f1, f2, ...]

def myload(output_of_mysave):
    return thing
```

wuhu:
Adapted the PR according to Friday's discussion:

The save function now does not take the `i` - the `path` already has it.
Instead, one can pass a file suffix as the third element of the tuple.

```python
def mysave(val, path):
    [ do the saving, return nothing ]

def myload(path):
    return [ load from path ]

x = value(x, (mysave, myload, '_append.this')
```

The complex case (see previous comment) is also supported.
I've added a section to the docs explaining both.

PR #209, opened 2021-11-04
Serialize all

wuhu:
#194


padl/dumptools/serialize.py
```diff
@@ -44,8 +46,20 @@ def __init__(self, val: Any, save_function: Callable, load_function: callable,
...
+        if self.file_suffix is not None:
+            path = Path(str(path) + f'/{self.i}{self.file_suffix}')
+        filename = self.save_function(self.val, path)
+        if filename is None:
+            assert self.file_suffix is not None, ('if no file file_suffix is passed to *value*, '
+                                                  'the *save*-function must return a filename')
+            filename = path.name
+        if isinstance(filename, (str, Path)):
+            complete_path = f"pathlib.Path(__file__).parent / '{filename}'"
+        else:
```
sjrl, line 58: Instead of using `else` could we use an `isinstance(filename, List)` and raise an exception if a different type is provided? This could catch an error in a custom written `save_function`

padl/dumptools/serialize.py
```diff
@@ -77,12 +91,11 @@ def save_all(cls, codegraph, scopemap, path):
                             scopemap[scoped_name] = SCOPE
```
sjrl, line 93: I think there is an extra new line

padl/transforms.py
```diff
@@ -1581,7 +1581,9 @@ def horizontal(n):
 
 class BuiltinTransform(AtomicTransform):
     def __init__(self, call):
-        super().__init__(call)
+        caller_frameinfo = inspector.non_init_caller_frameinfo()
```
sjrl, line 1584: Could you add a short docstring for this class? I also wanted to ask if why need this and `ClassTransform`?

sjrl:
Looks great! Could you also add a unit test for testing the custom saver with multiple file names?

padl/dumptools/serialize.py
wuhu: yes

Issue #192, opened 2021-11-03
save changing things `x = x + 1`


wuhu:
This
```
x = 1
x = x + 1
```
only saves
```
x = x + 1
```
right now. It should save both.
Same for `+=` etc.

To support this, we need to add a count.

wuhu:
#216

PR #216, opened 2021-11-08
Find reassignments

wuhu:
#192
Previously if a variable would be reassigned
```
a = 1
a = a + 1
```
the wouldn't be saved.
This PR fixes that.


sjrl:
Nice! Does this mean the following would work as well?
```python
from padl import transform
import numpy as np
np = transform(np)
```

wuhu:
@sjrl 
yes, that'll work.

Issue #142, opened 2021-10-30
enable saving as zip


wuhu:
Currently we save everything in a folder. This has the advantage to be transparent and simple. However, we don't end up with a single file.
It would be nice to be able to load and save as a compressed file.

PR #218, opened 2021-11-08
Enable saving as zip

wuhu:
#142


padl/transforms.py
```diff
@@ -165,13 +168,41 @@ def pd_post_load(self, path: Path, i: int):
         :param i: Unique transform index, can be used to construct filenames.
         """
 
+    def pd_zip_save(self, path: Union[Path, str], force_overwrite: bool = False):
+        """Save the transform to a zip-file at *path*.
+
+        The file's name should end with '.padl'. If no extension is given, it will be added
```
sjrl, line 174: Should we enforce that the suffix is `.padl`?

padl/transforms.py
```diff
@@ -1723,6 +1761,15 @@ def load(path):
     return transform
 
 
+def _zip_load(path: Union[Path, str]):
+    """Load a transform from a compressed '.padl' file. """
+    # we can't use a context because the files need to exist when using / saving again
```
sjrl, line 1766: What do you mean by this comment exactly? It looks like you are using a `with` context below?

padl/transforms.py
```diff
@@ -165,13 +168,41 @@ def pd_post_load(self, path: Path, i: int):
...
+
+        if path.exists():
+            if not force_overwrite:
+                raise FileExistsError(f'{path} exists, call with *force_overwrite* to overwrite.')
+            try:
+                rmtree(path)
+            except NotADirectoryError:
+                remove(path)
+
+        with TemporaryDirectory('.padl') as dirname:
```
sjrl, line 192: We should consider making the `TemporaryDirectory` inside of `path`. People may not have write permissions in other parts of the machine.

padl/transforms.py
wuhu: When using the context, the temporary directory is deleted when leaving. This doesn't work unfortunately as the files are needed for code inspection / saving. Thus we can't use the context.

sjrl: Could you add a note, specifying that the context we can't use refers to the `TemporaryDirectory`?

padl/transforms.py
```diff
@@ -1692,20 +1723,27 @@ def __call__(self, args):
...
+    if compress:
+        transform.pd_zip_save(path, force_overwrite)
+    else:
+        transform.pd_save(path, force_overwrite)
 
 
 def load(path):
     """Load a transform (as saved with padl.save) from *path*. """
+    if Path(path).is_file():
+        return zip_load(path)
```
sjrl, line 1746: Should this be calling `_zip_load`?

sjrl:
Could you add a unittest?

PR #228, opened 2021-11-11
Added tests for zip saving and variable renaming

sjrl:
@wuhu I modified `test_function_a` to use a variable rename `np = transform(np)` and it looks like the test hangs indefinitely when trying to save the transform. Could you take a look at this when you have the chance? 

- [x] Look at failing test


wuhu (approved):

wuhu:
@sjrl bug's fixed

sjrl:
Nice!

Issue #27, opened 2021-10-12
Implicit dataloader and its influence to the `__call__` method on a Transform and Subtransform.


eramdiaz:
Let us define a Transform like the following one:
```
t = (GPU(True)
      >> TracedLayer(torch.nn.Linear(8,4)) - 'dense_1'
      >> TracedLayer(torch.nn.Linear(4,2)) - 'dense_2'
)
```
It is definitely very nice how in `aleph` you can index and inspect the different Transforms in a composed Transform. 

So, if we inspect the third element (the linear layer `dense_2`) and call it, it could be thought that we are just computing the pass through a linear layer. This is the case if `t` is on infer mode. Then, you can do this:
```
t.infer()
inp_ = torch.randn(8)
output = t[2]( t[:2](inp_) )
```
where we are computing the pass of `inp_` through `GPU` and `dense_1` and then feeding explicitly calling `dense_2`. This feels natural and makes our Composed Transform look clear.

However, this won't work if the stage is `train` or `eval`. This is because in this case, there is an implicit Dataloader before making the layers passes: 
https://github.com/Aleph-One-GmbH/aleph/blob/b998f4f618e8cd22eeb608e23b3e35dbfbb9cd1a/aleph/transforms/core.py#L1692

Therefore, if we do something like this:
```
res = t[:2](torch.randn(6,8), batch_size=2, flatten=False, num_workers=0)
inp_ = list(res)
res2 = t[2](inp_, batch_size=2, flatten=False, num_workers=0)
list(res2)
```
`t[2]` won't just pass `inp_` thought `dense_2`. Before that another dataloader will be called with `inp_`, which is already batchified data. Then, an error will be raised.

I consider that the idea behind the previous codeblock is a desirable behaviour  and should be allowed by `lf`. Otherwise, users might find the structure of a composed Transform confusing and might need to go to the source code to understand it, which is something that might not be willing to.

Besides, it makes out Transforms consistent and has an intuitive concept behind. Also, I think that not being restricted to have a dataloader would lead to a more general framework.

One idea to acquire this would be including the dataloader call in a Transform. For example inside `GPU`, and not getting it by default in the `__call__` method of a Transform.

Issue #97, opened 2021-10-27
`lf_component` mapping and `preprocess, forward, postprocess` breakdown does not work in some edge cases


jasonkhadka:
Example:
```python
model_3 = (
    plus_one + times_two
    >> plus_one / Batchify()
    >> plus_one / plus_one
    >> Batchify() / times_two
    >> Unbatchify()
    >> plus_one
)
```

`pd_preprocess`
```python
(
    plus_one + times_two)
    >> (plus_one / lf.Batchify())
    >> (plus_one / plus_one)
    >> (lf.Batchify() / lf.Identity()
    )
```

`pd_forward`
```python
(lf.Identity() / times_two)
```

`pd_postprocess`
```python
(
    lf.Unbatchify() 
    >> plus_one
)
```

wuhu:
I've drafted a solution here: https://github.com/lf1-io/padl/pull/230

The idea is to implement a method `_pd_splits` for the different Transform classes which takes the component configuration of the input (i.e. information about which parts of the input are untouched, batchified, or unbatchified) and returns a 3-tuple of (preprocess, forward, postprocess)- parts and the output component configuration.
It seems that by having the precise input component configuration available everywhere, instead of using the `_pd_component` - set, makes it possible to solve the problem.

The above example works, it is split into:

Preprocess:
```
Compose:

      └──────────┐
      │          │
      ▼ x        ▼ x
   0: plus_one + times_two      
      │          │
      ▼ x        ▼ args
   1: plus_one / padl.Identity()
      │          │
      ▼ x        ▼ args
   2: plus_one / padl.Identity()
   ```

Forward:

```
Compose:

      └──────────┐
      │          │
      ▼ x        ▼ x
   0: plus_one + times_two      
      │          │
      ▼ x        ▼ args
   1: plus_one / padl.Identity()
      │          │
      ▼ x        ▼ args
   2: plus_one / padl.Identity()
```

Postprocess:

```
plus_one:

   def plus_one(x):
       return x + 1
```

PR #230, opened 2021-11-11
fix for split problem

wuhu:
- [x] Failing test for `IfInStage` when using `_pd_get_splits`
- [x] TODO Get this `(a / b / c) >> (x + y + z)` example to work. There is a test for this under `TestRollout`
- [x] TODO  Get this `(a / b) >> c` example to work. There is a test for this under `TestParallel`
- [x] Failing test for `TestTransformDeviceCheck`. Not sure if it's possible to get error `WrongDeviceError` to raise

- This solution reopens issue #27 since `_pd_get_splits()` relies on Transforms to be in preprocess by default. An alternative solution to this issue could be to directly use `transform.pd_call_transform()` when inspecting individual components of the transform. This would also avoid double batching. @wuhu @eramdiaz what do you think?


blythed (approved):

padl/transforms.py
```diff
@@ -93,6 +93,37 @@ def __init__(self, call_info: Optional[inspector.CallInfo] = None,
...
+        - *splits* is a 3-tuple of splits the entries are:
+            - the "preprocess" part of the transform
+            - the "forward" part of the transform
+            - the "postprocess" part of the transform
+        """
+        return (
+            # a normal transform doesn't change the components
+            input_components,
+            # for the component the transform is in, return the transform, else Identity
+            tuple(self if i == input_components else Identity() for i in range(3))
```
sjrl, line 124: Thanks for the additional explanation on this. Should the default transform have `input_components==1` since we put all Transforms into the `forward` split by default?

padl/transforms.py
```diff
@@ -93,6 +93,37 @@ def __init__(self, call_info: Optional[inspector.CallInfo] = None,
         self._pd_device = 'cpu'
         self._pd_layers = None
 
+    def _pd_splits(self, input_components=0) -> 'Transform':
```
sjrl, line 96: I'm not sure what the function signature should be, but I think something like `Tuple[Union[int, List], Tuple[Transform]]`

padl/transforms.py
```diff
@@ -1719,6 +1902,25 @@ def __init__(self, dim=0):
...
+
+    def _pd_splits(self, input_components=0) -> Transform:
+        """See the docstring of :meth:`Transform._pd_splits` for more details.
+
+        Batchify has empty splits and puts the component-number to 1 ("batchified").
+        """
+        # ensure that all inputs are "fresh"
+        assert self._all_0(input_components), 'double batchify'
+        # put the output component to 1 ("batchified") and return empty splits
+        return 1, (Identity(), Identity(), Identity())
```
sjrl, line 1922: `Batchify` should be returned and attached as part of the preprocess step

padl/transforms.py
```diff
@@ -1681,6 +1856,14 @@ def __init__(self, dim=0, cpu=True):
         self._pd_component = {'postprocess'}
         self.cpu = cpu
 
+    def _pd_splits(self, input_components=0) -> Transform:
+        """See the docstring of :meth:`Transform._pd_splits` for more details.
+
+        Unbatchify has empty splits and puts the component-number to 2 ("un-batchified").
+        """
+        # put the output component to 2 ("un-batchified") and return empty splits
+        return 2, (Identity(), Identity(), Identity())
```
sjrl, line 1865: `Unbatchify` should be returned in the postprocess

padl/transforms.py
wuhu: ah right, I think it should be `Tuple[Union[int, List], Tuple[Transform, Transform, Transform]]`

sjrl:
@wuhu Sorry I did a `force push` when rebasing with `main`. Just as a heads up

wuhu:
@sjrl nice! I also noticed the slowness, I think it's due to some overhead when creating `Identity` instances, we probably should reuse the same Identity everywhere.

padl/transforms.py
sjrl: Actually right now this would not be possible since `_pd_splits()` relies on Transforms to be in `preprocess` by default. This would bring back issue https://github.com/lf1-io/padl/issues/27.

An alternative solution to issue https://github.com/lf1-io/padl/issues/27 could be to directly use `transform.pd_call_transform()` when inspecting individual components of the transform. This would also avoid double batching. @wuhu @eramdiaz what do you think?

sjrl:
@wuhu I have sped things up somewhat by instantiating the `Identity` transform once and reusing it when calling `_pd_get_splits`. However, the tests still run somewhat slowly since `_pd_get_splits` is called multiple times and I am not caching. 

I tried to cache, but this can cause problems because if a transform is used in two separate Compose transforms it will "remember" what stage (preprocess, forward, postprocess) it was in before. 

```python
from padl import transform, Batchify, Identity

a = transform(lambda x: x + 1)
t1 = Identity() >> Batchify() >> a
t2 = a >> Batchify() >> Identity()

t1._pd_get_splits()  # a will be put in forward
t2._pd_get_splits()  # a will also be put into forward since it was cached
```

sjrl:
I got caching to work. If either `if self._pd_splits is None or self._pd_splits[0] != input_components:` then the splits are recalculated. This solves the issue I mentioned above.

sjrl:
Only one failing test `TestTransformDeviceCheck` with error message:
```python
    def test_device_check(self):
        self.transform_1.pd_to('gpu')
        self.transform_1.transforms[1].pd_to('cpu')
    
        with pytest.raises(WrongDeviceError):
>           self.transform_1.pd_forward_device_check()
E           Failed: DID NOT RAISE <class 'padl.exceptions.WrongDeviceError'>
```
With the changes I find it might be not possible to get the error to raise. This is because now every time `transform.pd_forward` is requested it will send everything in the `forward` stage to the `transform.pd_device`.

jasonkhadka:
https://github.com/lf1-io/padl/issues/97

jasonkhadka:
i tested the solution for some cases, and it looks like this solution works well.

sjrl:
> i tested the solution for some cases, and it looks like this solution works well.

If you have some particularly hard cases that didn't work before could you add them as tests? Or show me the code to make the transforms here and I'll add them as tests.

padl/transforms.py
```diff
@@ -91,9 +91,46 @@ def __init__(self, call_info: Optional[inspector.CallInfo] = None,
...
+
+        *input_components* contains information about the "split" the input is in. It is either an
+        int or a (potentially nested) list of ints. The ints have the following meaning:
+            - 0 means "not batchified"
+            - 1 means "batchified"
+            - 2 means "unbatchified" a.k.a. post-process
+        If it's a (nested) list, the structure represents the input structure for the transform,
+        whereas the entries represent the "split" of parts of the input.
+
+        # TODO When input_components is a list/tuple of ints does this indicate there are multiple "pipes"?
```
sjrl, line 110: @wuhu I'm trying to better understand `input_components`. So I would like to ask When input_components is a list/tuple of ints does this indicate there are multiple "pipes"?

padl/transforms.py
```diff
@@ -1490,6 +1500,64 @@ def __init__(self, transforms: Iterable[Transform], call_info: inspector.CallInf
...
+            Case 1:     Case 2:     Case 3:
+            pre + pre   Identity    Identity
+            for / for   for + for   Identity
+            pos / pos   pos / pos   pos + pos
+
+        The *output_components* are the list of output components of the sub-transforms.
+        """
+        if self._pd_splits is None or self._pd_splits[0][0] != input_components:
+            # TODO Should Rollout be able to accept input_components consisting of a list of ints?
+            assert isinstance(input_components, int), 'Rollout can only accept one input'
```
sjrl, line 1524: @wuhu Should Rollout be able to accept input_components consisting of a list of ints? I believe if it's a list then there are multiple "pipes" and `Rollout` is not valid in that case, right?

padl/util_transforms.py
```diff
@@ -52,29 +50,39 @@ def __call__(self, args):
...
+                for split, subsplit in zip(splits, subsplits):
+                    split.append(subsplit)
+
+            cleaned_splits = tuple(
+                builtin_identity if all(isinstance(s, Identity) for s in split) else split
+                for split in splits
+            )
+
+            final_splits = tuple(
+                IfInStage(
```
sjrl, line 77: @wuhu I'm not able to create a new instance of `IfInStage` in `_pd_get_splits`. I get the following error message when running `pytest -k TestIfInStage`
```python
    def test_infer_apply(self):
>       assert self.transform_1.infer_apply(1) == 9

tests/unittests/test_util_transforms.py:45: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
padl/transforms.py:734: in infer_apply
    self.pd_forward_device_check()
padl/transforms.py:1222: in pd_forward_device_check
    if isinstance(self.pd_forward, type(self)):
padl/transforms.py:639: in pd_forward
    forward = self._pd_get_splits()[1][1]
padl/transforms.py:1333: in _pd_get_splits
    (_, output_components), subsplits = transform_._pd_get_splits(output_components)
padl/util_transforms.py:76: in _pd_get_splits
    final_splits = tuple(
padl/util_transforms.py:77: in <genexpr>
    IfInStage(
padl/util_transforms.py:30: in __init__
    super().__init__(arguments=OrderedDict([('if_', if_),
padl/transforms.py:906: in __init__
    call = inspector.get_segment_from_frame(caller_frameinfo.frame, 'call')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

caller_frame = <frame at 0x7f9401eb26e0, file '/Users/sebastianlee/Aleph-One-Gmbh/padl/padl/util_transforms.py', line 77, code <genexpr>>, segment_type = 'call', return_locs = False

    def get_segment_from_frame(caller_frame: types.FrameType, segment_type, return_locs=False) -> str:
        """Get a segment of a given type from a frame.
    
        *NOTE*: All this is rather hacky and should be changed as soon as python 3.11 becomes widely
        available as then it will be possible to get column information from frames
        (see inline comments).
    
        *segement_type* can be 'call', 'attribute', 'getitem'.
        """
        if segment_type == 'call':
            node_type = ast.Call
            instructions_finder = _instructions_up_to_call
        elif segment_type == 'attribute':
            node_type = ast.Attribute
            instructions_finder = _instructions_in_name
        elif segment_type == 'getitem':
            node_type = ast.Subscript
            instructions_finder = _instructions_in_getitem
        # we want to extract the precise init statement here (e.g. `MyClass(1, 2, 3)`
        # , for python 3.11 (currently in development) this can be done via co_positions
        # (see https://www.python.org/dev/peps/pep-0657/),
        # for now, as 3.11 isn't widely used, this requires the following hack:
        # extract the source of the class init statement
        try:
            full_source = caller_frame.f_globals['_pd_source']
        except KeyError:
            full_source = get_source(caller_frame.f_code.co_filename)
    
        source, offset = get_statement(original(full_source), caller_frame.f_lineno)
        # the source can contain surrounding stuff we need to discard
        # as we only have the line number (this is what makes this complicated)
    
        # get all segments in the source that correspond to calls and might thus
        # potentially be the class init
        candidate_segments = var2mod.Finder(node_type).get_source_segments(source)
        # for each candidate, disassemble and compare the instructions to what we
        # actually have, a match means this is the correct statement
        if not candidate_segments:
            raise RuntimeError('No attributes found.')
        # disassemble and get the instructions up to the current position
        target_instrs = _instructions_up_to_offset(caller_frame.f_code,
                                                   caller_frame.f_lasti)
    
        segment = None
        locs = None
        found = False
    
        for segment, locs in sorted(candidate_segments, key=lambda x: -len(x[0])):
    
            instrs = instructions_finder(segment)
            if len(instrs) > len(target_instrs):
                continue
            for instr, target_instr in zip(instrs, target_instrs[-len(instrs):]):
                if (target_instr.opname == 'LOAD_FAST'
                        and target_instr.argval in caller_frame.f_locals
                        and target_instr.argval in caller_frame.f_code.co_varnames):
                    argval = caller_frame.f_locals[target_instr.argval]
                else:
                    argval = target_instr.argval
                if instr.argval != target_instr.argval and instr.argval != argval:
                    break
                same_opname = instr.opname == target_instr.opname
                load_ops = ('LOAD_NAME', 'LOAD_FAST', 'LOAD_GLOBAL', 'LOAD_CONST')
                both_load = instr.opname in load_ops and target_instr.opname in load_ops
                if not (same_opname or both_load):
                    break
            else:
                found = True
                break
    
        if segment is None or not found:
>           raise RuntimeError(f'{segment_type} not found.')
E           RuntimeError: call not found.

```

padl/transforms.py
wuhu: Yes, if `input_components` is a list, it indicates multiple pipes. 
However, if it's not a list, it could still be multiple pipes, all in the same "component".

padl/transforms.py
wuhu: It's still valid, it just means that those multiple pipes get rolled out.

For example, this
```
(a / b / c) >> (x + y + z)
```
would cause the Rollout `(x + y + z)` to get a list [0, 0, 0] (or other numbers, depending on what happens in a, b, c). This is because the output of (a / b / c) - the input of (x, y, z) has three parts. That input would be passed to each of the three transforms a, b and c. Thus, the whole list would be passed to all three, x, y and z.

padl/util_transforms.py
wuhu: I'll look into that. Seems like it's the same problem we had before with returning Identity (this happens if a Transform inside transforms.py creates another transfor), which I solved by introducing the BuiltinTransform. I'll check if there's some better solution to that.

wuhu:
> Only one failing test `TestTransformDeviceCheck` with error message:
> 
> ```python
>     def test_device_check(self):
>         self.transform_1.pd_to('gpu')
>         self.transform_1.transforms[1].pd_to('cpu')
>     
>         with pytest.raises(WrongDeviceError):
> >           self.transform_1.pd_forward_device_check()
> E           Failed: DID NOT RAISE <class 'padl.exceptions.WrongDeviceError'>
> ```
> 
> With the changes I find it might be not possible to get the error to raise. This is because now every time `transform.pd_forward` is requested it will send everything in the `forward` stage to the `transform.pd_device`.

Does this mean that we can remove the test? Or is it a bug?

sjrl:
> > Only one failing test `TestTransformDeviceCheck` with error message:
> > ```python
> >     def test_device_check(self):
> >         self.transform_1.pd_to('gpu')
> >         self.transform_1.transforms[1].pd_to('cpu')
> >     
> >         with pytest.raises(WrongDeviceError):
> > >           self.transform_1.pd_forward_device_check()
> > E           Failed: DID NOT RAISE <class 'padl.exceptions.WrongDeviceError'>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With the changes I find it might be not possible to get the error to raise. This is because now every time `transform.pd_forward` is requested it will send everything in the `forward` stage to the `transform.pd_device`.
> 
> Does this mean that we can remove the test? Or is it a bug?

It's not a bug, so I would say we could remove the test. What do you think @jasonkhadka ?

padl/util_transforms.py
sjrl: It doesn't seem to be a problem with the `CompoundTransform`'s

padl/transforms.py
sjrl: Okay I understand. I'll remove the `assert` and add a test for this case, to make sure things work as expected.

padl/transforms.py
sjrl: I'll remove the TODO and add this info into the doc string.

jasonkhadka:
> > > Only one failing test `TestTransformDeviceCheck` with error message:
> > > ```python
> > >     def test_device_check(self):
> > >         self.transform_1.pd_to('gpu')
> > >         self.transform_1.transforms[1].pd_to('cpu')
> > >     
> > >         with pytest.raises(WrongDeviceError):
> > > >           self.transform_1.pd_forward_device_check()
> > > E           Failed: DID NOT RAISE <class 'padl.exceptions.WrongDeviceError'>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > With the changes I find it might be not possible to get the error to raise. This is because now every time `transform.pd_forward` is requested it will send everything in the `forward` stage to the `transform.pd_device`.
> > 
> > 
> > Does this mean that we can remove the test? Or is it a bug?
> 
> It's not a bug, so I would say we could remove the test. What do you think @jasonkhadka ?

The assertion can just be changed to 
```
assert self.transform_1.pd_forward_device_check()
```
instead of 
```
with pytest.raises ...
```

To test if the forward is always anyway in correct device when called, even when part of forward has been sent to wrong device.

wuhu:
@sjrl 

> This solution reopens issue Implicit dataloader and its influence to the __call__ method on a Transform and Subtransform. #27 since _pd_get_splits() relies on Transforms to be in preprocess by default. An alternative solution to this issue could be to directly use transform.pd_call_transform() when inspecting individual components of the transform. This would also avoid double batching. @wuhu @eramdiaz what do you think?

Hmm I think this might make it a bit harder to understand - the user would need to know that and it's not very transparent why that is the case.

Would it work if we, while performing `_pd_get_splits`, check whether there is any batchify and, if that's not the case, move the 0-split (pre-batchified) to 1 (batchified)?

sjrl:
> @sjrl
> 
> > This solution reopens issue Implicit dataloader and its influence to the **call** method on a Transform and Subtransform. #27 since _pd_get_splits() relies on Transforms to be in preprocess by default. An alternative solution to this issue could be to directly use transform.pd_call_transform() when inspecting individual components of the transform. This would also avoid double batching. @wuhu @eramdiaz what do you think?
> 
> Hmm I think this might make it a bit harder to understand - the user would need to know that and it's not very transparent why that is the case.
> 
> Would it work if we, while performing `_pd_get_splits`, check whether there is any batchify and, if that's not the case, move the 0-split (pre-batchified) to 1 (batchified)?

I'm not entirely sure how to implement this, but it probably could be done.

wuhu:
@sjrl I've fixed the tests related to rollout and parallel.
The problem with
```python
        request.cls.transform_4 = (
            plus_one / plus_one
            >> transform(lambda x: x[0] * x[1])
        )
```
was that the default `Transform._pd_get_splits` expected an int for `input_component` and would return all `Identities` if that wasn't the case (as happened here for the `lambda`) I changed it such that the input can be a nested list and added an assertion that all ints in that list are the same.

The failing `Rollout` test was correct actually (both `times_two`s get a tuple, multiplying a tuple repeats the items instead of multiplying them). I've changed the expected output.

sjrl:
> > Would it work if we, while performing _pd_get_splits, check whether there is any batchify and, if that's not the case, move the 0-split (pre-batchified) to 1 (batchified)?
> 
> @sjrl if you don't mind I would try and come up with a solution for this. While on it I'd also make two changes to the caching:
> 
> * cache using a dict where the keys are the input_component
> * move common code (caching and things related to the split moving) to a method in Transform like you did previously with `pd_forward`, `_pd_forward_part` etc
> 
> what do you think?

@wuhu Both of those points sound good to me!

wuhu:
I've made some changes, all is now working

- I've changed it such that only Transform has a _pd_get_stages whereas I renamed the previous _pd_get_stages to _pd_splits. _pd_get_stages caches the result of _pd_splits and rearranges the splits if no Batchify is present. Thus "splits" means "pre-batchify", "post-batchify", "post-unbatchify" whereas "stages" means the usual, "preprocess", "forward", "postprocess".
- I've fixed the IfInstage and added a check that the output components of the "if" and "else" transform match.
- some tests didn't match the changed logic, I've changed that
- I noticed a failing print, fixed it and added tests

blythed (commented):
Seems fine.

jasonkhadka:
i see a problem :

```python
import padl
from padl import transform, batch, unbatch, identity

@transform
def plus_one(x):
    return x + 1


@transform
def plus(x, y):
    return x + y


test = (
    plus_one
    >> batch
    >> identity + identity
    >> plus
    >> unbatch
    >> identity + identity
    >> plus
)
```

gives:
```python
> test.pd_preprocess
Compose:

      │
      ▼ x
   0: plus_one       
      │
      ▼ args
   1: Batchify(dim=0)

> test.pd_forward
Compose:
      │
      ▼ x
   0: plus_one       
      │
      ▼ args
   1: Batchify(dim=0)

> test.pd_postprocess
Compose:

      │
      ▼ args
   0: Unbatchify(dim=0, cpu=True)
      │
      ▼ (x, y)
   1: plus      
```

`pd_splits` is removing the identity rollouts. This bug should be fixed.

jasonkhadka (changes requested):
commented above regarding unwanted removal of identity.

wuhu:
@jasonkhadka fixed that

jasonkhadka:
> @jasonkhadka I fixed that
great! I just have a small comment on assertion, else this looks good.

jasonkhadka (approved):

padl/transforms.py
```diff
@@ -135,9 +135,81 @@ def __init__(self, call_info: Optional[inspector.CallInfo] = None,
...
+        - *output_components* is the "splits" information of the output, it has the same format as
+        the *input_components*.
+
+        - *splits* is a 3-tuple of splits, the entries are:
+            - the "pre-batchified" part of the transform
+            - the "batchified" part of the transform
+            - the "postprocess" part of the transform
+        """
+        component = Transform._pd_merge_components(input_components)
+        assert isinstance(component, int)
```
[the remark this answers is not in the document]
wuhu: :+1:

PR #249, opened 2021-11-15
Bug 246

### LINKED ISSUE #246: Saving things referencing themself is broken.
Saving things referencing themself:

Minimum not-working examples:

(a) setting an attribute
```python
@transform
class T:
    def __init__(self):
        self.a = 1
        self.b = self.a  # problem
    
    def __call__(self):
        ...
```

(b) setting at an index
```python
@transform
def f():
    x = [1, 2, 3]
    x[0] = x[1]
```

Both is due to a missing check in `dumptools.var2mod._VarFinder.visit_Attribute` (it treats this as an assignment to `self` (a) or `x` (b)).

wuhu:
## Description

Fixes #246 \<issue_number>


sjrl (approved):
This looks great!

blythed (approved):

PR #251, opened 2021-11-15
Bug 241

### LINKED ISSUE #241: `value` is ignored
Saving code:

```python
f = open('checkpoints/imagenet/index_to_name.json')
LABELS = value(json.load(f))
f.close()
```

Serialized code:

```python
f = open('checkpoints/imagenet/index_to_name.json')
LABELS = value(json.load(f))
```

This is incorrect, the `value` should mean that the `LABELS` variable is loaded from the saved directory.

wuhu:
## Description

Fixes #241 \<issue_number>

### DIFF
--- a/tests/material/save_by_value.py
+++ b/tests/material/save_by_value.py
@@ -1,0 +1,10 @@
+from padl import transform, value, save
+
+x = [1, 2, 3]
+LABELS = value(x)
+
+
+@transform
+def test(x):
+        return LABELS[x]
+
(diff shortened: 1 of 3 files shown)


tests/material/save_by_value.py
```diff
@@ -0,0 +1,10 @@
+from padl import transform, value, save
+
+x = [1, 2, 3]
+LABELS = value(x)
+
+
+@transform
+def test(x):
+        return LABELS[x]
```
sjrl, line 9: Looks like there is an over indent for `return LABELS[x]`

PR #252, opened 2021-11-15
Make batchify act more like the pytorch DataLoader

sjrl:
## Description

I propose to modify `Batchify` to act more like the pytorch data loader in that it will return a `torch.tensor` of `torch.tensor`'s instead of a `tuple` of `torch.tensor`'s. This would only affect `Batchify` when calling with the `infer` stage. It does not break any of the current tests and should work the same (more or less) as returning a `tuple`. 

Most importantly, I think this makes the behavior more consistent between `infer` and `eval`/`train` where after the `Batchify` step you can be certain you will always be working with a `torch.tensor`.

What do you think @wuhu @blythed @jasonkhadka @eramdiaz ?

### DIFF
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -1730,7 +1730,7 @@ class Batchify(ClassTransform):
         if Transform.pd_stage != 'infer':
             return args
         if isinstance(args, (tuple, list)):
-            return tuple([self(x) for x in args])
+            return torch.tensor([self(x) for x in args])
         if isinstance(args, torch.Tensor):
             return args.unsqueeze(self.dim)
         if isinstance(args, (float, int)):


wuhu:
I just tried with the `DataLoader` and in my case it returned a (nested) list of tensors:

```python
from torch.utils.data import DataLoader
import torch

class DataSet:
    def __getitem__(self, item):
        return (torch.tensor([1,1,1]), 
                (torch.tensor([2,2,2]), torch.tensor([3,3,3])))
    
    def __len__(self):
        return 1000

x = DataLoader(DataSet(), batch_size=3)
next(iter(x))
```
returns
```
[tensor([[1, 1, 1],
         [1, 1, 1],
         [1, 1, 1]]),
 [tensor([[2, 2, 2],
          [2, 2, 2],
          [2, 2, 2]]),
  tensor([[3, 3, 3],
          [3, 3, 3],
          [3, 3, 3]])]]
```
- when does it return a tensor of tensors for you?

sjrl:
Ahh okay I see. You are right, I think I was getting confused with our expected behavior between `infer` and `eval`/`train`. The behvior is already consistent expect that the DataLoader seems to always return a list instead of tuples.

PR #253, opened 2021-11-15
Have _move_to_device return the same type as was given

sjrl:
## Description

I think `_move_to_device` should only move torch tensors to device and not modify the structure of the args. So if a `tuple` of `tensors` is given it returns a `tuple` of `tensors` and the same for `lists` of `tensors`. Right now it will convert a `list` of `tensors` to a `tuple` of `tensors` which could be potentially confusing.


wuhu:
:+1: makes sense

wuhu (approved):

PR #259, opened 2021-11-16
Fix test for Try

sjrl:
@eramdiaz These changes will allow your tests to pass. I realize that this does not resolve the issue of saving a Transform that has been defined as an attribute of a class, but that can be resolved in a different PR from this one I think. What do you think?


eramdiaz:
Yes, I agree it could be resolved in a different pr since it is an independent issue. Thanks @sjrl

PR #261, opened 2021-11-16
Try

eramdiaz:
## Description

Try Transform and tests implemented.


padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+    :param transform: Transform to try
+    :param catch_transform: Transform to fall back on
+    :param exceptions: Catch conditions
+    """
+
+    def __init__(self,
+                 transform: Transform,
+                 catch_transform: Transform,
+                 exceptions: Union[List, Tuple, Exception],
+                 else_transform: Transform = Identity(),
```
sjrl, line 126: add doc string for `else_trasnform`

padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+                 finally_transform: Transform = Identity(),
```
sjrl, line 127: add doc string for `finally_transform`

padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+                 pd_name: str = None):
```
sjrl, line 128: Add docstring for `pd_name`

padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+
+        if not isinstance(exceptions, (tuple, list)):
+            exceptions = (exceptions, )
+        exceptions = tuple(exceptions)
+        for exception in exceptions:
+            assert issubclass(exception, Exception)
+        super().__init__(pd_name=pd_name,
+                         arguments=OrderedDict([('transform', transform),
+                                                ('catch_transform', catch_transform),
+                                                ('exceptions', exceptions)]))
```
sjrl, line 138: Should the `else_transform` and `finally_transform` be added to the `OrderedDict`?

padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+        self.transform = transform
+        self.catch_transform = catch_transform
+        self._exceptions = exceptions
```
sjrl, line 141: I don't think you need to make the `exceptions` private. It's possible users may want to access that quantity.

padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+        self.else_transform = else_transform
+        self.finally_transform = finally_transform
+        self._pd_component = set.union(self.transform._pd_component,
```
sjrl, line 144: Should the `union` statement include `self.catch_transform`?

padl/util_transforms.py
```diff
@@ -108,3 +108,51 @@ class IfTrain(IfInMode):
...
+                                       self.else_transform._pd_component,
+                                       self.finally_transform._pd_component)
+
+        assert len(self._pd_component) == 1, 'Stage must not change inside a Try Transform.'
```
sjrl, line 148: Right now we are using the word `component` to describe `preprocess`, `forward`, etc. So please change `Stage` to `Component` for now.

sjrl: I would also consider rephrasing the error message to `Try transform cannot contain transforms that have multiple components`

tests/unittests/test_util_transforms.py
```diff
@@ -80,3 +101,75 @@ def test_save_and_load_3(self, tmp_path):
         self.transform_3.pd_save(tmp_path / 'test.padl')
         t3 = padl.load(tmp_path / 'test.padl')
         assert t3.infer_apply(1) == 12
+
+
+class TestTry:
+    @pytest.fixture(autouse=True, scope='class')
+    def init(self, request):
+        request.cls.to_int = transform(lambda x: 1)
```
sjrl, line 109: Remove `to_int` if it's not used

tests/unittests/test_util_transforms.py
```diff
@@ -80,3 +101,75 @@ def test_save_and_load_3(self, tmp_path):
...
+        request.cls.str_sum = transform(lambda x: x + 'a')
```
sjrl, line 110: Remove `str_sum` if it's not used

tests/unittests/test_util_transforms.py
```diff
@@ -80,3 +101,75 @@ def test_save_and_load_3(self, tmp_path):
...
+        request.cls.return_zero = transform(lambda x: 0)
```
sjrl, line 111: Remove `return_zero` if it is not used

padl/util_transforms.py
sjrl: Would you prefer to keep this private?

padl/util_transforms.py
```diff
@@ -108,3 +108,58 @@ class IfTrain(IfInMode):
...
+class Try(ClassTransform):
+    """ Perform *transform*. If this fails with any exception from *exceptions*, perform
+    *catch_transform*.
+
+    :param transform: Transform to try.
+    :param catch_transform: Transform to fall back on.
+    :param exceptions: Catch conditions.
+    :param else_transform: Transform to carry on the `else` clause of the `try` statement.
+    :param finally_transform: Transform to carry on on the `finally` clause of the `try` statement.
+    :param pd_name:
```
sjrl, line 122: You can find an example doc string for this variable in `transforms.py`

padl/util_transforms.py
eramdiaz: ok, let's make it public then, thanks @sjrl

padl/util_transforms.py
```diff
@@ -108,3 +108,58 @@ class IfTrain(IfInMode):
...
+                                       self.else_transform._pd_component,
+                                       self.finally_transform._pd_component)
+
+        assert len(self._pd_component) == 1, 'Try Transform cannot contain transforms that have ' \
+                                             'multiple components.'
+
+    def __call__(self, args):
+        try:
+            args = self.transform.pd_call_transform(args)
+        except self._exceptions:
```
sjrl, line 160: Update `self._exceptions` to `self.exceptions`

padl/util_transforms.py
```diff
@@ -108,3 +108,58 @@ class IfTrain(IfInMode):
...
+
+    def __call__(self, args):
+        try:
+            args = self.transform.pd_call_transform(args)
+        except self.exceptions:
+            args = self.catch_transform.pd_call_transform(args)
+        else:
+            args = self.else_transform.pd_call_transform(args)
+        finally:
+            return self.finally_transform.pd_call_transform(args)
```
wuhu, line 165: I'm wondering:
What would happen if an exception which isn't in self.exceptions occurs within the try block?
In that case the finally block would get the input "args", rather than the args which come out of self.transform.pd_call_transform. Those would be used if no exception happens.
Thus we have two conflicting cases - maybe we should make finally always use the input args or disallow finally. @eramdiaz what do you think?

padl/util_transforms.py
```diff
@@ -108,3 +108,58 @@ class IfTrain(IfInMode):
 
     def __init__(self, if_: Transform, else_: Optional[Transform] = None):
         super().__init__(if_, 'train', else_)
+
+
+class Try(ClassTransform):
+    """ Perform *transform*. If this fails with any exception from *exceptions*, perform
+    *catch_transform*.
```
wuhu, line 115: you could also explain the else and finally cases here

eramdiaz: Right

padl/util_transforms.py
eramdiaz: Let me know if I understood correctly, but I believe this is desirable behaviour in a `try` clause. If I do:
```
def g(x):
    return x/0

def f(x):
    try:
        x = g(x)
    except SyntaxError:
        x = x + 1
    else:
        x = x + 2
    finally:
        return x
 f(5) # This returns 5
 ```
I don't want whatever comes out from `g` to go to `finally` but `x`, in this case, 5.

Comment (wuhu) [reply padl/util_transforms.py:165]: Here's an example of what I mean:

```python
def g(x):
    return str(1 / x)

def f(x):
    try:
        x = g(x)
    except TypeError:
        x = 'failed'
    finally:
        return x + ' bla bla bla'
```
if I put a `1` I would get `'1 bla bla bla'`. If I put a `"hello"`, I would get 'failed bla bla bla' because "hello" causes a `TypeError` in `g`. If, however, I put `0`, `g(x)` would fail with a `ZeroDivisionError` and I would go directly to `finally`. Because x didn't pass through `g`, it is still an int and I would get a `TypeError`.

So what I mean is we can't be sure of what kind of thing the "finally_transform" would get, which could lead to unexpected behaviour.

eramdiaz: I understand we could have some unexpected behaviours, but I consider that should be under the user's responsability, who is the one that define all the `Transform` executed in the `Try` `Transform`. 

If someone wants to include a `finally` transform and we don't include it on `Try`, then the user would have to write an independent `Transform` with `try`, `except`, `else` and `finally`, which makes our `Try` transform much less useful.

sjrl: In that case I think instead of making `else` and `finally`, the `Identity()` transform by default they should be `None`. This would mean if they aren't specified then they are not included in the block. For example, 
```python
t = Try(a, b, None, None)

# The call method of t should look like
__call__(self, args):
   try:
       return a(args)
   except:
       return b(args)
```
So we don't require the user to always have to specify `else` and `finally`.

wuhu: I think you are right that it would be nice to keep the finally in - that way we make sure that the users can do everything they can with a normal try block.

I don't agree however that we should keep it the way that it is currently. The problem is that you are asking the user to make the `finally_transform` such that it will always work correctly with both the `args` from the input and the transformed `args`. This would make it hard to build a correct `finally_transform` (sometimes it might be impossible if there is no way to tell which `args` it is) and it's also hard to understand / explain why it should be that way. I would not be surprised if most users would get this wrong.

How about the following instead:

```python
        [...]
        try:
            output = self.transform.pd_call_transform(args)
        except self.exceptions:
            output = self.catch_transform.pd_call_transform(args)
        else:
            output = self.else_transform.pd_call_transform(output)
        finally:
            self.finally_transform.pd_call_transform(args)
        return output
```
that way it would always be clear what the input of the finally is. I also don't think we're losing much generality because if the user wants to do something with the output in every case they can always just put a transform after the Try.

wuhu: Here's an example of something that would work well the way I proposed in the previous comment (a bit silly to do this exactly of course, but I think it's a good example of what kind of thing finally is usually used for):

```python
@transform
def open_file(filename):
    return open(filename)

@transform
def read(file):
    return file.read()

@transform
def close(file):
    file.close()
 
read_from_path = (
     open
     >> Try(read, finally=close)
)
```

eramdiaz: Ok, I agree it would be better without the `else` and `finally`transforms

wuhu (approved):
Nice! merging.

PR #267, opened 2021-11-17
unbatchify fix

blythed:
## Description

Fixes #\<issue_number>


padl/transforms.py
```diff
@@ -565,14 +606,24 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
...
 
             if flatten:
+
                 if verbose:
                     pbar.update()
-                if not use_post:
-                    output = Unbatchify(cpu=False)(batch)
+
+                # TODO this was edited
+                if use_post:
```
sjrl, line 618: `post.pd_call_transform` is now hidden under the `flatten` flag. Was this intended?

padl/transforms.py
```diff
@@ -1798,3 +1851,22 @@ def group(transform: Union[Rollout, Parallel]):
     2-tuple whose first item will be passed to `a` and whose second item will be passed to `b + c`.
     """
     return transform.grouped()
+
+
+class _ItemGetter:
```
sjrl, line 1856: Add back the docs from the previous method
```python
    """A simple dataset.
    Takes *samples* and applies *transform* to it.
    :param samples: An object implementing __getitem__ and __len__.
    :param transform: Preprocessing transform.
    :param exception: Exception to catch for (fall back to *default*).
    :param default: The default value to fall back to in case of exception.
    """
```

padl/transforms.py
```diff
@@ -565,14 +606,24 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
...
+                # output = post.pd_call_transform(output, mode)
 
             if flatten:
+
                 if verbose:
                     pbar.update()
-                if not use_post:
-                    output = Unbatchify(cpu=False)(batch)
+
+                # TODO this was edited
```
sjrl, line 617: Remove TODO

padl/transforms.py
sjrl: This means the entire post step could be skipped if the user doesn't specify `flatten=True`

sjrl: Sorry, I didn't see the `assert` above.

blythed: Yes all intended

PR #271, opened 2021-11-17
Bug recursion

wuhu:
## Description
Recursive transforms like 

```python
@transform
def f(x):
    if x == 0:
        return x
    return x + f(x - 1)
```

would not save. This fixes it.

+ some refactoring and extra documentation.


sjrl:
Gotta love recursion! Nice!

sjrl (approved):

PR #274, opened 2021-11-18
Refactor

wuhu:
Some cleaning up and extra docstrings.

### DIFF
--- a/tests/unittests/test_dump.py
+++ b/tests/unittests/test_dump.py
@@ -1,6 +1,7 @@
 import pytest
 
 from padl import transform, group, IfTrain, Batchify, Unbatchify
+from tests.material import transforms_in_module as tim
 
 
 CONST = 1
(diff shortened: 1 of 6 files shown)


tests/unittests/test_dump.py
```diff
@@ -232,3 +233,9 @@ def test_dumping_indexset():
 
 def test_dumping_selfassign():
     assert SelfAssign(1)._pd_dumps() == read_dump('selfassign')
+
+
+def test_othermodule_function():
+    x = tim.function._pd_dumps()
+    # TODO: wip
+    #  assert 'todo' == 'done'
```
sjrl, line 241: Should this assert be uncommented? Or left as is for moment?

sjrl (approved):
Looks great!

tests/unittests/test_dump.py
wuhu: No it should be left as it is.

Issue #275, opened 2021-11-18
Clean up named_tuple logic in call of compoud transform


jasonkhadka:
* clean up named_tuple logic, so that final output is always named tuple for Parrallel/Rollout
* Add tests for this logic

sjrl:
Issue #75 is related to this.
- Call method for `Unbatchify` and `Batchify` should return `namedtuple` if the input is a `namedtuple`

Issue #276, opened 2021-11-18
Replace namedtuple


jasonkhadka:
## Description
Namedtuple has some issues like it cannot be pickled causing problems in dataloading, and applying transform. If this can be replaced with something that is better suited with purpose (somethign between dictionary and tuple), call logic can be simplified. 

## Solution
New datastructure or new tuple class that can be pickled.

## Alternatives
Current method, that causes several problem:
* Cannot be used in dataloader where num_workers>1
* This causes issue that inside apply tuples are passed instead of namedtuple, hence, the get by field name cannot be used inside transform

sjrl:
Issue #75 is related to this.

PR #281, opened 2021-11-18
Fix named tuple

jasonkhadka:
## Description

Fixes https://github.com/lf1-io/padl/issues/275

   * clean up named_tuple logic, so that final output is always named tuple for Parrallel/Rollout
   * Output is namedtuple for all *apply methods in all modes.
   * Add tests for this logic


wuhu:
is this related to #276?

tests/unittests/test_transforms.py
```diff
@@ -129,6 +129,53 @@ def test_isinstance_of_namedtuple():
...
+        assert not pd._isinstance_of_namedtuple(self.transform_1.infer_apply(1))
+        assert pd._isinstance_of_namedtuple(self.transform_2.infer_apply(1))
+        assert pd._isinstance_of_namedtuple(self.transform_3.infer_apply(1))
+
+    def test_eval_apply(self):
+        assert not any(list(map(pd._isinstance_of_namedtuple,
+                                self.transform_1.eval_apply([1, 2, 3])
+                                )
+                            )
+                       )
```
sjrl, line 149: Could we put all of these `)` on the same line to save some space. So `)))`

sjrl:
@jasonkhadka Shouldn't we also remove 
```python
        if Transform.pd_mode is not None:
            return tuple(out)
```
from the call methods in `Rollout` and `Parallel` for the time being?

sjrl:
> is this related to #276?

Issue #276 will be a larger refactor so we can used `namedtuple`'s in between transforms. Right now a `namedtuple` can only be returned as the final output of the `_apply` methods. We cannot used `namedtuple`'s everywhere because they cannot be pickled which makes them incompatible with the `DataLoader`.

wuhu:
> Issue #276 will be a larger refactor so we can used namedtuple's in between transforms. Right now a namedtuple can only be returned as the final output of the _apply methods. We cannot used namedtuple's everywhere because they cannot be pickled which makes them incompatible with the DataLoader.

I see, thanks for the explanation.

jasonkhadka:
> @jasonkhadka Shouldn't we also remove
> 
> ```python
>         if Transform.pd_mode is not None:
>             return tuple(out)
> ```
> 
> from the call methods in `Rollout` and `Parallel` for the time being?

this is still needed cause users might still want to call single transform and expect named output. 
example:
```python
t = (a - 'a1' + b - 'b1') >> (c / d)
```
users might do `t[0](arg)` and expect output as `namedtuple('a1': a(arg), 'b1':b(arg))` 
This logic makes it possible.

sjrl (approved):
Looks good to me!

Issue #268, opened 2021-11-17
look into CI/ CD with github actions


blythed:

sjrl:
- action to run tests on PR
- action to recompile docs?

jasonkhadka:
* [x]  action to run tests on PR
* [ ]  cache dependencies
* [ ] linting
* [ ]  action to recompile docs?


>     * [ ]  action to run tests on PR
> 
>     * [ ]  action to recompile docs?

Issue #205, opened 2021-11-04
add more detailed documentation


wuhu:
- [x] operators
- [x] stages
- [x] preprocess / forward / postprocess
- [x] this
- [x] serializers

jasonkhadka:
also make the examples in doc work. 
`pytest --doctest-modules`  fails at the moment.

sjrl:
> _if needed_ also make the examples in doc work. `pytest --doctest-modules` fails at the moment.

I've started going through the doctest examples to see if I can get them to pass in branch https://github.com/lf1-io/padl/tree/doc_tests

blythed:
Examples for the frontpage website:

- Side-by-side preprocess/ dataloader old way vs. padl
- loading model old way and new way
- ...

jasonkhadka:
building model old way 

```python
class SubNet1(torch.nn.Module):
    def forward(self, x):
          ...

class SubNet2(torch.nn.Module):
    def forward(self, x):
          ...

class SuperNet(torch.nn.Module):
    def __init__(self):
            self.subnet1 = SubNet1()
            self.subnet2 = SubNet2()
    def forward(self, x):
          x = self.subnet1(x)
          x = self.subnet2(x)
          return x
```

new way:


```python
@transform
class SubNet1(torch.nn.Module):
    def forward(self, x):
          ...

@transform
class SubNet2(torch.nn.Module):
    def forward(self, x):
          ...

SuperNet = (
     SubNet1
     >> SubNet2
)
```

wuhu:
Hmm I think that's cheating a bit, you could just do

```python
SuperNet = torch.nn.Sequential(
    SubNet1(),
    SubNet2()
)
```

PR #282, opened 2021-11-19
Ci/Cd for padl

jasonkhadka:
## Description

Fixes https://github.com/lf1-io/padl/issues/268

Adds pytesting for PRs. 

In next rounds:
- linting
- type checks


.github/workflows/actions.yml
```diff
@@ -0,0 +1,36 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r requirements-test.txt
+    - name: Test with pytest
+      run: |
+        python -m pytest --cov=padl -n 2
```
sjrl, line 36: Does the coverage get logged somewhere?

requirements.txt
```diff
@@ -2,3 +2,4 @@ torch
 tqdm
 numpy
 astunparse
+pytest
```
sjrl, line 5: I don't think `pytest` needs to be in here right?

.github/workflows/actions.yml
sjrl: I just found it under the actions tab: https://github.com/lf1-io/padl/runs/4262619217?check_suite_focus=true

sjrl:
For future rounds, it might be worth looking into caching dependencies to speed up our tests: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs-or-python?langId=py#caching-dependencies

.github/workflows/actions.yml
jasonkhadka: yeah, it is printed on the logs of test step, but not stored in file.

sjrl:
In issue #268 I also brought up the possibility of using an action to rebuild the docs. Would that be possible? I think currently we run the risk of having our docs be out of date with the current source code.

requirements.txt
jasonkhadka: yeah true, i was testing there and forgot to remove it.

jasonkhadka:
yeah, i think those could be done. I will add them for next updates to actions. 
For now, we dont have a big list of dependencies, so it is farily quick to run the workflow.

> In issue #268 I also brought up the possibility of using an action to rebuild the docs. Would that be possible? I think currently we run the risk of having our docs be out of date with the current source code.

sjrl:
> yeah, i think those could be done. I will add them for next updates to actions. For now, we dont have a big list of dependencies, so it is farily quick to run the workflow.

Do you know if we have "unlimited" computation time for these actions?

sjrl:
@jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?

jasonkhadka:
> > yeah, i think those could be done. I will add them for next updates to actions. For now, we dont have a big list of dependencies, so it is farily quick to run the workflow.
> 
> Do you know if we have "unlimited" computation time for these actions?

yes, seems to be unlimited for public repos.

sjrl:
> @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?

Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.

jasonkhadka:
> > @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?
> 
> Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.

yeah, dont think this is needed right now. Might be considered while re-writing docs.

wuhu:
> > @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?
> 
> Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.

Agree to that, many of the examples are not written with doctest in mind (rather they are just for illustration), for example they make reference to things that don't exist like this:

```
>>> mytransform(mysamplething)
blu
```

jasonkhadka:
> > > @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?
> > 
> > 
> > Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.
> 
> Agree to that, many of the examples are not written with doctest in mind (rather they are just for illustration), for example they make reference to things that don't exist like this:
> 
> ```
> >>> mytransform(mysamplething)
> blu
> ```

just added a small comment in issue #205 (docs related) to think about it, but yeah, would make examples writing bit more complicated.

sjrl:
> > > @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?
> > 
> > 
> > Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.
> 
> Agree to that, many of the examples are not written with doctest in mind (rather they are just for illustration), for example they make reference to things that don't exist like this:
> 
> ```
> >>> mytransform(mysamplething)
> blu
> ```

@wuhu @jasonkhadka I actually think it might be worth getting these to pass. I'm already going through the doctests now to see if I can get them to pass. I've been making good progress and I may have caught a bug in one of the `dumptools.sourceget` functions.

jasonkhadka:
I think its better to take this discussion on the issue regarding documentation : https://github.com/lf1-io/padl/issues/205
we can list all the problems of examples, and perhaps discuss on writing new examples for those there. 

> > > > @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?
> > > 
> > > 
> > > Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.
> > 
> > 
> > Agree to that, many of the examples are not written with doctest in mind (rather they are just for illustration), for example they make reference to things that don't exist like this:
> > ```
> > >>> mytransform(mysamplething)
> > blu
> > ```
> 
> @wuhu @jasonkhadka I actually think it might be worth getting these to pass. I'm already going through the doctests now to see if I can get them to pass. I've been making good progress and I may have caught a bug in one of the `dumptools.sourceget` functions.



> > > > @jasonkhadka Also it may be worth adding `pytest --doctest-modules` to our testing to make sure all examples in our docstrings pass. What do you think?
> > > 
> > > 
> > > Ehh I'm not sure about this at the moment. A lot of them don't pass and I'm not sure if we should spend the effort to get them to work right now.
> > 
> > 
> > Agree to that, many of the examples are not written with doctest in mind (rather they are just for illustration), for example they make reference to things that don't exist like this:
> > ```
> > >>> mytransform(mysamplething)
> > blu
> > ```
> 
> @wuhu @jasonkhadka I actually think it might be worth getting these to pass. I'm already going through the doctests now to see if I can get them to pass. I've been making good progress and I may have caught a bug in one of the `dumptools.sourceget` functions.

sjrl (approved):
Looks good to me!

wuhu (commented):
neat!

Issue #160, opened 2021-11-01
Allow saving whole transform code if it was created in another module


wuhu:
If I get a transform from another module and do

```python
from othermodule import MyTransform
x = MyTransform()
```

saving will save just that - it will not get the code defining the transform.

Sometimes it could be desirable to save the code defining the Transform instead.

We should somehow enable that.

Different ways come into mind:
1. make it possible to set a flag in the decorator

```python
@transform(saveall=True)
class MyTransform:
    ...   
```

2. add a function that modifies the way the transform saves

```python
from othermodule import MyTransform
x = MyTransform()
x = saveall(x)
```

3. add a function that modifies the way a whole package / module saves

```python
import othermodule
saveall(othermodule)
```

blythed:
For example in `tatenen` models should be saved with reference to `tatenen` codebase.

wuhu:
https://github.com/lf1-io/padl/pull/288

sjrl:
@wuhu Can this issue be closed now?

PR #288, opened 2021-11-21
Enable choosing between full dump or import dump for external transforms 

wuhu:
#160 

- [x] class transform
- [x] function transform
- [x] torch-module transform
- [x] compound transform
- [x] tests
- [x] toggle for transforms
- [x] toggle for modules / packages
- [x] documentation
- [x] prevent name conflicts


padl/dumptools/sourceget.py
```diff
@@ -28,11 +28,20 @@ def get_source(filename: str, use_replace_cache: bool = True) -> str:
     If *use_replace_cache*, try getting the source from the "replace_cache", which can contain
     explicit replacements of the original source strings.
     """
+    # needed to support doctest
+    if filename.startswith('<doctest'):
```
sjrl, line 32: Nice! Great to see that `doctest` had a solution.

sjrl:
It looks like all `doctest-modules` pass. I think it would be a good idea to add the option `--doctest-modules` to the `pytest` command in the `.github/workflows/actions.yml` file. What do you think? @wuhu

padl/dumptools/sourceget.py
sjrl: @wuhu Could you point me to a good resource for learning about these different caching mechanisms in `python`? I've had a hard time finding a good explanation for how to navigate these different cases.

padl/transforms.py
```diff
@@ -123,21 +115,42 @@ class Transform:
...
+
+    @property
+    def _pd_full_dump(self) -> bool:
+        """If *True*, dump the Transform in full (with definition etc) even if it was defined in
+        a different module. Else, only dump an import statement. """
+        module = self._pd_full_dump_relevant_module
+        # always fully dump Transforms from the module the dump was triggered in
+        if inspector.caller_module() == module:
+            return True
+        # fully dump all Transforms from packages or mudules specified in
```
sjrl, line 148: Spelling error `mudules` -> `modules`

padl/transforms.py
```diff
@@ -352,21 +365,36 @@ def pd_save(self, path: Union[Path, str], force_overwrite: bool = False):
         with open(path / 'versions.txt', 'w') as f:
             f.write(versions)
 
-    def _pd_codegraph_startnode(self, name: str) -> var2mod.CodeNode:
-        """Build the start-code-node - the node with the source needed to create *self* as "name".
-        (in the scope where *self* was originally created). """
+    def _pd_codegraph_add_startnodes(self, graph, name: Union[str, None]) -> CodeNode:
+        """Build the start-:class:`CodeNode` objecs - the node with the source needed to create
```
sjrl, line 369: I think `objecs` -> `objects`?

padl/transforms.py
```diff
@@ -411,19 +439,12 @@ def f(x):
         The codegraph can be used to compile a python module defining ``f``.
 
         :param graph: A codegraph to extend. If *None* a new codegraph will be created.
-        :param scopemap: A dict mapping scoped names to the scopes they were created in.
         :param name: The name to give the transform.
         :param scope: The scope for the start-node. Default is to use the scope of the transform.
```
sjrl, line 443: Remove `:param scope:` since it is no longer passed as an argument.

padl/transforms.py
```diff
@@ -352,21 +365,36 @@ def pd_save(self, path: Union[Path, str], force_overwrite: bool = False):
         with open(path / 'versions.txt', 'w') as f:
             f.write(versions)
 
-    def _pd_codegraph_startnode(self, name: str) -> var2mod.CodeNode:
-        """Build the start-code-node - the node with the source needed to create *self* as "name".
-        (in the scope where *self* was originally created). """
+    def _pd_codegraph_add_startnodes(self, graph, name: Union[str, None]) -> CodeNode:
```
sjrl, line 368: I think the return type of this function should be changed to `Set`

sjrl (approved):
Looks great! Thanks for all of the additional documentation and examples.

padl/dumptools/sourceget.py
wuhu: I don't have one but it's this:
The frameinfo (see https://docs.python.org/3/library/inspect.html) object has a filename field. This allows us to get the source of what's currently being executed.
For a normal python module it's the filename.
For ipython and doctest the code is not in files, therefore it's different.
For ipython, every cell has an id and the "filename" contains a string with that id. The content of the code can than be found in linecache.cache (https://docs.python.org/3/library/linecache.html).
For doctest neither works, the code in the docstrings is just executed. We can get the code from the fileinfo.code_context which the context of the code being executed - this is not perfect as it contains only a few lines, but for doctest, it looks like it's enough.

wuhu:
@sjrl doctests fail in 3.9 because some objects have different names, I suggest to fix that in a different PR (created an issue at #318).

PR #305, opened 2021-12-08
Adding github action for building docs

sjrl:
## Description

Added github action to automatically build the sphinx docs and commit the changes to the PR. If no changes are detected (the docs don't change) then no commit is made.


docs/apidocs/padl.html
```diff
@@ -195,290 +142,63 @@
...
-    
   </ul>
-
-  
   <hr/>
 </div>
           <div role="main" class="document" itemscope="itemscope" itemtype="http://schema.org/Article">
            <div itemprop="articleBody">
-            
-  <div class="section" id="module-padl">
```
sjrl, line 263: It looks like the API docs are not being regenerated. I'm guessing there is a problem with the Makefile. @wuhu @blythed Do you have an idea of what is going wrong?

sjrl: This is fixed in the new PR.

PR #309, opened 2021-12-09
Adding macos and windows to CI/CD

sjrl:
## Description

Adding MacOS and Windows to the CI/CD workflow in addition to ubuntu


docs/py-modindex.html
```diff
@@ -0,0 +1,286 @@
+<!DOCTYPE html>
```
jasonkhadka, line 1: is this html file needed?

jasonkhadka (changes requested):
there are 2 failed tests and not sure if the html file is needed in pr.

sjrl:
@wuhu @jasonkhadka @blythed I also added tests for the Windows OS to see what would happen. It looks like we fail the Windows + 3.9. I wanted to ask if we ever plan on supporting Windows? If not I can remove them.

docs/py-modindex.html
sjrl: Yes this file was generated when building the docs and was missed in the last PR.

jasonkhadka:
i think we should defenitely support windows. 
Also @sjrl  
You can run the changes to this CI/CD if you add the name of this branch in actions.yml
on:
  push:
    branches:
      - main

sjrl:
> i think we should definitely support windows. Also @sjrl You can run the changes to this CI/CD if you add the name of this branch in actions.yml
> 
> ```
> on:
>   push:
>     branches:
>       - main
> ```

I think that is already present in `actions.yml` right? Or am I missing something?

jasonkhadka:
> > i think we should definitely support windows. Also @sjrl You can run the changes to this CI/CD if you add the name of this branch in actions.yml
> > ```
> > on:
> >   push:
> >     branches:
> >       - main
> > ```
> 
> I think that is already present in `actions.yml` right? Or am I missing something?

yes, it is alright, sorry wrong suggestion here.

sjrl:
We decided to leave this open until the Windows + 3.9 CI build is fixed. @wuhu @eramdiaz

wuhu:
fixed in #330 and merged into there, closing here

PR #313, opened 2021-12-10
Enable star imports for modules without __all__.

wuhu:
Previously saving things depending on star imports would only work with modules defining `__all__`. This fixes it.


sjrl (approved):
Looks great!

PR #315, opened 2021-12-13
Changes to make padl python 3.7 compatible

sjrl:
## Description

Some changes needed to make PADL compatible 3.7


wuhu:
Main problems were 
- missing `ast.get_source_segment`
- missing `ast.AST.end_lineno` and `ast.AST.end_col_offset`

Fixed by using the `asttokens` library if needed in new module `dumptools/ast_utils.py`.

wuhu:
- [ ] can we live with the slowness? (tests ~3 times slower, only affects creating and loading transforms, not execution)

.github/workflows/build_docs.yml
```diff
@@ -46,5 +46,6 @@ jobs:
       run: |
         git config --global user.name 'LF1 Bot'
         git config --global user.email 'lf1_bot@users.noreply.github.com'
+        git add .
```
sjrl, line 49: To be a little safer could we do `git add docs/`? I'm afraid with the current call we could accidentally files we don't mean to.

sjrl:
> * [ ]  can we live with the slowness? (tests ~3 times slower, only affects creating and loading transforms, not execution)

Is this slowness only occur when `python 3.7` is used? If so, I'd say better to have the support than not even if it is slower in building and loading.

sjrl:
@wuhu I also ran through one of our notebooks and I did not notice the slowness when creating the transforms.

sjrl:
> 

I've continued to play around with padl in a notebook and I am beginning to think the slowness is linked to running in the `pytest` environment. When running `test_pd_splits_rollout` using `pytest -k test_pd_splits_rollout --durations=0` the test takes about 9 sec to run vs 1 sec normally. However, if I copy this test over into a jupyter notebook running python 3.7 and run it there it takes less than 1 sec.

.github/workflows/build_docs.yml
wuhu: yes, makes sense, I'll change

wuhu:
> Is this slowness only occur when python 3.7 is used? If so, I'd say better to have the support than not even if it is slower in building and loading.

Yes, the way I refactored it everything is basically the same as before for python >= 3.8.

> I've continued to play around with padl in a notebook and I am beginning to think the slowness is linked to running in the pytest environment. When running test_pd_splits_rollout using pytest -k test_pd_splits_rollout --durations=0 the test takes about 9 sec to run vs 1 sec normally. However, if I copy this test over into a jupyter notebook running python 3.7 and run it there it takes less than 1 sec.

My understanding is that the larger the files the transforms come from, the slower this is. Seems like for finding out where statements end asttokens goes through the whole file and counts the characters, basically. This makes `test_transforms` slow as the module is rather long.

So it could become a problem for larger codebases (like tenen), but notebooks I should be fine. Also the critical code is isolated in `dumptools/ast_utils.py` thus if we find something more efficient we can easily replace.

sjrl:
> So it could become a problem for larger codebases (like tenen), but notebooks I should be fine. Also the critical code is isolated in `dumptools/ast_utils.py` thus if we find something more efficient we can easily replace.

I see, that makes sense. I think it's fine to merge then. We could consider adding a warning to the docs about this when using 3.7.

wuhu:
> I see, that makes sense. I think it's fine to merge then. We could consider adding a warning to the docs about this when using 3.7.

Agree.

@blythed, what do you think?

Summary is: 
- we can make it work for 3.7
- this ads an extra dependency for 3.7 only ("asttokens")
- it's relatively slow if transforms come from large files (also, only 3.7 no difference for other versions)
- in notebooks basically no difference

Should we use this or not use this or look for a more efficient fix?

Issue #316, opened 2021-12-13
remove tqdm dependency


wuhu:
## 🛰️ Feature

Currently we only use `tqdm` in `_pd_itercall` when `verbose=True`. This isn't really necessary as instead of

```python
for batch in model.eval_transform(l, verbose=True):
    ...
```
one can do

```python
for batch in tqdm(model.eval_transform(l)):
    ...
```

which in my mind is a bit nicer because 
- more explicit
- no need for extra requirement
- we leave the choice of progress bar to the user

(we might need to do a little change to _pd_itercall in order to make it get the length properly).

sjrl:
I agree this makes it more explicit. Additionally, users may want to toggle between `from tqdm.notebook import tqdm` and `from tqdm import tqdm` depending on what environment they are in.

PR #325, opened 2021-12-14
#316 remove tqdm

wuhu:
Removing tqdm.

#316

Output of `_pd_itercall` is now a `_GeneratorWithLength` instance. This is to give it a length such that when calling

```python
for x in tqdm(t.eval_apply(...)):
    ...
```

we get a progress bar instead of just a counter.


padl/transforms.py
```diff
@@ -713,39 +725,29 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
...
-                    output_format = self._pd_get_output_format()
-                    if output_format is not None:
-                        yield output_format(*out)
-                    else:
-                        yield out
-            else:
-                output_format = self._pd_get_output_format()
-                if output_format is not None:
-                    yield output_format(*output)
+        def _gen():
```
sjrl, line 728: Thanks for simplifying the logic here!

jasonkhadka:
great! 
Should an example also be included in documentation to show how `tqdm` should be used along with `train_apply` and `eval_apply`?

sjrl:
> 

I think that would be a great idea! Also we can update the notebook examples to use `tqdm` to show that as well. Speaking of which @wuhu can you check that the `verbose` option was not used in the notebooks and that it does not appear anywhere else in the docs?

wuhu:
@jasonkhadka @sjrl thanks for the reviews! I've now double checked that we don't use verbose anywhere and added a section to the docs showing how to use it with tqdm.

jasonkhadka (approved):

sjrl (approved):

PR #336, opened 2021-12-16
Update setup.py

sjrl:
## Description

Update description in setup.py


sjrl:
@wuhu I'm not sure why the Automated Doc Builder got triggered here. Could you take a look?

Issue #304, opened 2021-12-07
Grouping not kept in `pd_preprocess`, `pd_forward` and `pd_postprocess`


eramdiaz:
When we have a Transform `g` compounded by several elements and any of them is grouped, this grouping is not respected in `g.pd_preprocess` `g.pd_forward`, `g.pd_postprocess`. For example,

Let `t`  and `g` be:
```
linear = transform(torch.nn.Linear)(8, 4)
t = ((to_tensor >> padl.Batchify() >>  linear)
     + (to_tensor >> padl.Batchify() >> linear)
) - 'name' 
g = t + padl.Identity()
```
`t` is:
``` 
Rollout - "name":

   ├─▶ 0: to_tensor >> Batchify(dim=0) >> <anonymous Linear>
   │ +  
   └─▶ 1: to_tensor >> Batchify(dim=0) >> <anonymous Linear>
```
`g` is:
```
Rollout:

   ├─▶ 0: group([..>>..] + [..>>..])
   │ +  
   └─▶ 1: padl.Identity()
```
And `g.pd_preprocess` and `g.pd_forward` are, respectively:
```
Rollout:

   ├─▶ 0: to_tensor >> Batchify(dim=0)
   │ +  
   ├─▶ 1: to_tensor >> Batchify(dim=0)
   │ +  
   └─▶ 2: padl.Identity()
```
,
```
Parallel:

   ││└─▶ 0: Linear(in_features=8, out_features=4)
   ││  /  
   │└──▶ 1: Linear(in_features=8, out_features=4)
   │   /  
   └───▶ 2: padl.Identity()
```
@sjrl and me have been checking this. Maybe it would be more consistent if the grouping were kept on the `pd_forward` and `pd_preprocess`, so they would be, respectively, something like
```
Rollout:

   ├─▶ 0: group(...)
   │ +  
   ├─▶ 1: padl.Identity()
```
and
```
Parallel:

   │└─▶ 0: group(...)
   │  /  
   └──▶ 1: padl.Identity()
```

Update: this also happens in Rollouts of Parallels and viceversa, for example in
```
t = ((to_tensor >> batch >> linear) / (to_tensor >> batch >> linear)
     + (to_tensor >> batch) / (to_tensor >> batch)
)
```
.
where `t.pd_preprocess` and `t._pd_forward` are, respectively,
```
Rollout:

   ├─▶ 0: [..>>..] / [..>>..]
   │ +  
[...]

PR #338, opened 2021-12-20
Name the splits of named transforms.

wuhu:
solves #304

### DIFF
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -1552,6 +1552,13 @@ class Pipeline(Transform):
             deduped_keys.append(new_name)
         return deduped_keys
 
+    def _add_name_to_splits(self, final_splits):
+        """Ad name to split-transforms. """
+        if self._pd_name is not None:
+            for i, s in enumerate(final_splits):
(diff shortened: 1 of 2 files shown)


padl/transforms.py
```diff
@@ -1837,8 +1842,6 @@ def _pd_splits(self, input_components=0) -> Tuple[Union[int, List],
             else:
                 final_splits.append(s)
 
-        final_splits = tuple(final_splits)
```
sjrl, line 1840: Do we want to return the `final_splits` as a tuple?

padl/transforms.py
```diff
@@ -1623,6 +1623,11 @@ def _pd_splits(self, input_components=0, has_batchify=False) -> Tuple[Union[int,
             else:  # if it's empty: identity
                 final_splits.append(identity)
 
+        if self._pd_name is not None:
```
sjrl, line 1626: It looks like this code block is the same for the three cases. Could we make it a member function of `Pipeline` or `Transform`?

eramdiaz:
solved in the pr #338

PR #339, opened 2021-12-21
Debugger 3

eramdiaz:
## Description
Implementation of a customized debugger for padl transforms.


padl/transforms.py
```diff
@@ -37,6 +40,16 @@
     make_green, create_arrow, format_argument, visible_len
 
 
+class _Notset:
+    # pylint: disable=too-few-public-methods
+    def __bool__(self):
+        return False
+
+
+_notset = _Notset()
```
sjrl, line 49: I believe `_notset` was removed by @wuhu in an earlier PR. I don't think this is needed anymore, right @wuhu ?

padl/transforms.py
```diff
@@ -131,9 +144,9 @@ def __len__(self):
 class Transform:
     """Transform base class.
 
-    :param call_info: A `CallInfo` object containing information about the how the transform was
-        created (needed for saving).
-    :param pd_name: name of the transform
+    :param call_info: A `CallInfo` object containing information about the how the transform wasv
+    created (needed for saving).
```
sjrl, line 148: Please re-add the indent.

sjrl: And change `wasv` to `was`

padl/transforms.py
```diff
@@ -564,19 +595,21 @@ def _pd_dumps(self, return_versions: bool = False,
     def __repr__(self):
         return self._pd_shortrepr(formatting=False)
 
-    def _repr_pretty_(self, p, cycle) -> str:
-        # pylint: disable=invalid-name
+    def _repr_pretty_(self, p, cycle):
```
sjrl, line 598: Re-add the `pylint: disable ...`

padl/transforms.py
```diff
@@ -672,6 +705,15 @@ def _pd_unpack_argument(self, arg) -> bool:
             return True
         return False
 
+    def _pd_call_transform(self, arg):
```
sjrl, line 708: I think we should come up with a better name for this function. I find it a little confusing that `pd_call_transform` then calls `_pd_call_transform`.

padl/transforms.py
```diff
@@ -710,6 +750,9 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
         """
         assert mode in ('eval', 'train'), '_pd_itercall can only be used with mode eval or train'
 
+        global _pd_trace
+        _pd_trace.clear()
```
sjrl, line 754: I'm not too familiar with how globals in python work, but is this the right place to declare `_pd_trace`? This is also needed for `infer_apply` which doesn't call `_pd_itercall`.

padl/transforms.py
```diff
@@ -723,20 +766,29 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
         if use_preprocess:
             loader = self.pd_get_loader(args, preprocess, mode, **loader_kwargs)
         else:
-            loader = args
+            loader = list(zip(args, range(len(args))))
 
         def _gen():
-            for batch in loader:
+            for batch, ix in loader:
```
sjrl, line 772: Could we use `enumerate` here instead of using `list` and `zip` above?

padl/transforms.py
```diff
@@ -723,20 +766,29 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
...
                 batch = _move_to_device(batch, self.pd_device)
-
                 output = batch
                 if use_forward:
-                    output = forward.pd_call_transform(batch, mode)
+                    try:
+                        output = forward.pd_call_transform(batch, mode)
+                    except Exception as err:
+                        self._pd_trace_error(self._pd_get_error_idx('forward'),
+                                             [args[i] for i in ix])
```
sjrl, line 780: I thought `ix` refers to the batch idx when you are using the Pytorch dataloader. Is it actually a list of indices?

padl/transforms.py
```diff
@@ -931,6 +997,16 @@ def _pd_direct_subtransforms(self) -> Iterator[Transform]:
             if isinstance(v, Transform):
                 yield v
 
+    def _pd_get_non_target_stage_idx(self, stage_used):
```
sjrl, line 1000: I would appreciate documentation for these three functions `_pd_get_non_target_stage_idx`, `_pd_get_target_stage_idx`, and `_pd_get_error_idx` even though they are hidden. It's not clear to me what they are for.

padl/transforms.py
```diff
@@ -997,11 +1073,15 @@ def _pd_get_signature(self) -> List[str]:
...
         try:
-            return '\n'.join(self.source.split('\n')[:30])
+            str_ = self.source.split('\n')[:30]
+            if marker:
+                return str_[0] + marker[1] + '\n'.join(str_[1:])
+            return '\n'.join(str_)
         except TypeError:
-            return self._pd_call
+            return self._pd_call + marker[1] + '\n' if marker \
+                else self._pd_call
```
sjrl, line 1084: I think you can put this all on one line.

padl/transforms.py
```diff
@@ -1372,6 +1459,11 @@ def __getitem__(self, item: Union[int, slice, str]) -> Transform:
     def __len__(self):
         return len(self.transforms)
 
+    def _pd_call_transform(self, arg):
```
sjrl, line 1462: Why does `Map` need to have its own `_pd_call_transform`? I think we would ideally only define this function once for the base `Transform` class.

padl/transforms.py
```diff
@@ -1552,16 +1644,36 @@ def _pd_get_keys(transforms):
             deduped_keys.append(new_name)
         return deduped_keys
 
+    def _pd_get_non_target_stage_idx(self, stage_used):
```
sjrl, line 1647: Add docs

padl/transforms.py
```diff
@@ -1552,16 +1644,36 @@ def _pd_get_keys(transforms):
...
+        return 1
+
+    def _pd_get_target_stage_idx(self, is_entire_transform=None):
```
sjrl, line 1650: Add docs

padl/transforms.py
```diff
@@ -1552,16 +1644,36 @@ def _pd_get_keys(transforms):
...
+        """
+        assert stage in ('forward', 'postprocess')
+        return _pd_trace[-1].error_position
+
 
 class Compose(Pipeline):
     """Apply series of transforms on input.
 
-    Compose([t1, t2, t3])(x) = t3(t1(t2(x)))
+    Compose([t1, t2, t3])(x) = t3(t2(t1(x)))
```
sjrl, line 1671: Nice catch!

padl/transforms.py
```diff
@@ -1641,17 +1759,16 @@ def _pd_classify_nodetype(i, t, t_m1, cw, cw_m1):
...
+    def _pd_longrepr(self, formatting=True, marker=None) -> str:  # TODO: make it respect the formatting
         """Create a detailed formatted representation of the transform. For multi-line inputs
         the lines are connected with arrows indicating data flow.
         """
         # pad the components of rows which are shorter than other parts in same column
         rows = [
             [s._pd_parensrepr() for s in t.transforms] if hasattr(t, 'transforms')
             else [t._pd_shortrepr()]
-            for t in self.transforms
+            for i, t in enumerate(self.transforms)
```
sjrl, line 1770: I don't think the `enumerate` is needed here. Could you change it back?

padl/transforms.py
```diff
@@ -1743,10 +1861,56 @@ def __call__(self, args):
...
         :return: Output from series of transforms.
         """
-        for transform_ in self.transforms:
-            args = transform_.pd_call_transform(args)
+        _in_args = args
+        for i, transform_ in enumerate(self.transforms):
+            try:
+                args = transform_.pd_call_transform(args)
+            except Exception as err:
+                self._pd_trace_error(i, _in_args)
```
sjrl, line 1869: @eramdiaz What are the `_in_args` used for in the traceback?

padl/transforms.py
```diff
@@ -1743,10 +1861,56 @@ def __call__(self, args):
...
+            t.pd_forward = (t_3 >> t_4) + (t_5 >> t_6).
+            No matter what branch is failing on :meth:`t.pd_forward`, the error on `t` is on
+            len(t.pd_preprocess) + 0.
+        """
+        assert stage in ('forward', 'postprocess')
+        preprocess = self.pd_preprocess
+        forward = self.pd_forward
+        postprocess = self.pd_postprocess
+        preprocess_used = False if (isinstance(preprocess, Identity)) and \
+                                   not isinstance(self.__getitem__(0), Identity) else True
```
sjrl, line 1901: Why can't we use the same check for `preprocess_used` as in `_pd_itercall`?

padl/transforms.py
```diff
@@ -1743,10 +1861,56 @@ def __call__(self, args):
...
+
+        preprocess_idx = preprocess._pd_get_non_target_stage_idx(preprocess_used)
+        if stage == 'forward':
+            is_entire_transform = str(self) == str(forward)
```
sjrl, line 1905: I feel like comparing the strings of the transforms to check they are the same is very fragile. 

 We should check with @wuhu if we can reimplement the comparison operator for two transforms.

padl/transforms.py
```diff
@@ -1847,6 +2009,11 @@ def _pd_splits(self, input_components=0) -> Tuple[Union[int, List],
                 res.append(split)
         final_splits = res
 
+        if self._pd_name is not None:
+            for i, s in enumerate(final_splits):
+                if not isinstance(s, Identity):
+                    final_splits[i] = s - self._pd_name
+
```
sjrl, line 2016: Re-add `final_splits = tuple(final_splits)` here. We want to always return the final_splits as tuples.

padl/transforms.py
eramdiaz: yeah, makes sense, I might got confused when solving merging conflicts

padl/transforms.py
```diff
@@ -861,11 +913,25 @@ def infer_apply(self, inputs):
...
+            inputs = self.pd_forward.pd_call_transform(inputs, mode='infer')
+        except Exception as err:
+            self._pd_trace_error(self._pd_get_error_idx('forward'), in_args)
+            raise err
+        try:
+            inputs = self.pd_postprocess.pd_call_transform(inputs, mode='infer')
+        except Exception as err:
+            self._pd_trace_error(self._pd_get_error_idx('postprocess'), in_args)
+            raise err
+        output_format = self._pd_get_output_format()
```
sjrl, line 930: Could you change the last few lines to `return self._pd_format_output(inputs)`? This is causing a lot of the tests to fail currently.

padl/transforms.py
eramdiaz: done

padl/transforms.py
eramdiaz: done ^^

sjrl:
- [ ] Please add unit tests for the debugger

padl/transforms.py
eramdiaz: mmm yes, what about `_pd_run_with_try_if_needed`? not a very nice name either

padl/transforms.py
eramdiaz: `_pd_trace` is also defined on the line 51

padl/transforms.py
sjrl: - [ ] Also consider adding types to the arguments

padl/transforms.py
eramdiaz: yes, that's better

padl/transforms.py
```diff
@@ -1552,16 +1644,36 @@ def _pd_get_keys(transforms):
...
         return deduped_keys
 
+    def _pd_get_non_target_stage_idx(self, stage_used):
+        return 1
+
+    def _pd_get_target_stage_idx(self, is_entire_transform=None):
+        return 0
+
+    def _pd_get_error_idx(self, stage: str):
+        """Track the index where a `Pipeline` fails from the one that fails on `self.pd_forward`.
```
sjrl, line 1654: When referring to methods in docs please use 
```python
:meth:`Pipeline`
:meth:`self.pd_forward`
```
This allows sphinx to link to the methods in the docs.

padl/transforms.py
eramdiaz: Yes, `_ItemGetter` has been changed to retrieve also indices.

sjrl: Does that mean `ix` can be a batch of index numbers?

padl/transforms.py
sjrl: Ahh I see. Is there a reason you declare it as `global` here, but not on line 51?

padl/transforms.py
eramdiaz: I wanted to make sure that the `args` we are keeping on the trace are the first ones that are input to the `Compose` and not intermediate ones, but it might be not needed

padl/transforms.py
eramdiaz: got you, but it is normally the way we do it. but it would be nice to have a stronger way yeah

padl/transforms.py
eramdiaz: It is not enough with `use_preprocess`, `use_forward` on the debugger is that in order to track properly where the error occurs when looking at the entire transform (not at the different stages). I explain it with a strange but allowed example. Let `t` be
```
t = prep >> batch >> unbatch >> post_1
```
and `t2` be:
```
t2 = prep >> batch >> Identity() >> unbatch >> post_1
```
and suppose an exception is got in `post_1`. When we are trying to track when the error happens considering the entire transform, we compute this:
```
len(t.pd_preprocess) + AtomicTransform()._pd_get_non_target_stage_idx() + index_failing_on_postprocess
```
Equation 1.
For both of them, `use_forward` is `True` because the statement for `use_forward` is:
```
use_forward = not isinstance(forward, Identity)
```
but one of them has to add a 1 on the second summand of  Equation 1 and the second has to add a 0

padl/transforms.py
eramdiaz: ok

padl/transforms.py
eramdiaz: True

padl/transforms.py
sjrl: I see, even if it is the way we do it now, maybe it's worth putting this into the overloaded comparison operator `==` or as a separate function. This will make it easier to update in the future.

padl/transforms.py
sjrl: actually I'm checking with @wuhu if this is needed.

padl/transforms.py
sjrl: I see. I'm not sure I'm a huge fan of this because it could potentially cause a lot of overhead since we are keeping a copy of the input args. For example, if the input args happen to be a large matrix we are doubling the memory overhead.

padl/transforms.py
eramdiaz: This `_pd_call_transform` belongs to `Pipeline`. It is because `Pipeline` objects already trace the errors on its `__call__`

padl/transforms.py
eramdiaz: right

padl/transforms.py
sjrl: Do you mean that `use_forward` is `False` for both?

padl/transforms.py
eramdiaz: well, the problem with this is i need each element with the following order: (args, item) and enumerate gives it to me as (item, args)

padl/transforms.py
sjrl: I see. Is it possible to keep the `try` and `except` only in the `_pd_call_transform` and out of the `__call__` methods? This would reduce the number of places that we would need the `try` and `except` boiler plate.

padl/transforms.py
sjrl: I think I see how this is different. But I don't think this check will always work. For example 
```python
import padl
from padl import transform, batch, unbatch, Identity
f = transform(lambda x: x + 1)
t = padl.Identity() >> f
preprocess_used = False if (isinstance(preprocess, Identity)) and not isinstance(t[0], Identity) else True
# prints True even though there is no preprocessing step
```

padl/transforms.py
sjrl: Why do you need it in the order (args, item)?

sjrl: A potential issue I see with using `list` is that if someone passed in `args` as an `iterator`, similar to our `_ItemGetter` is that it would force the `iterator` to be completely moved into memory. This could then cause problems if the `iterator` is too big to fit into memory.

padl/transforms.py
eramdiaz: > Do you mean that `use_forward` is `False` for both?

yes

padl/transforms.py
eramdiaz: thanks! nice catch, i think i changed that recently by mistake

padl/transforms.py
eramdiaz: As I understand the `global` keyword is for changing the value of a global variable inside a function, but I might not know everything about it

padl/transforms.py
eramdiaz: > Why do you need it in the order (args, item)?

yes, I did not want to change so much the code but it is a better call, I changed the order to `for ix, batch` in loader (and also the corresponding line in `_ItemGetter`)

eramdiaz: > A potential issue I see with using `list` is that if someone passed in `args` as an `iterator`, similar to our `_ItemGetter` is that it would force the `iterator` to be completely moved into memory. This could then cause problems if the `iterator` is too big to fit into memory.

I see, then it would be nice to use an iterator, but the problem with `enumerate` is that it does not have a `__len__` method, but we could use a customized one with that method, what do you think?

padl/transforms.py
eramdiaz: yes, it is a batch of index numbers

padl/transforms.py
eramdiaz: ok, I'll do this

padl/transforms.py
eramdiaz: I think not, it is needed to do it inside the `__call__` of the `Pipeline`s because I need to trace the item of the `Pipeline` that is getting the Exception.

padl/transforms.py
eramdiaz: I'll keep this in mind and try to find an alternative

padl/transforms.py
eramdiaz: yeah, I am missing some of them

padl/transforms.py
eramdiaz: > I think I see how this is different. But I don't think this check will always work. For example
> 
> ```python
> import padl
> from padl import transform, batch, unbatch, Identity
> f = transform(lambda x: x + 1)
> t = padl.Identity() >> f
> preprocess_used = False if (isinstance(preprocess, Identity)) and not isinstance(t[0], Identity) else True
> # prints True even though there is no preprocessing step
> ```

right, so then maybe I should not consider this case and include this trivial `Identity`s on the ones considered on this issue: https://github.com/lf1-io/padl/issues/340

padl/transforms.py
sjrl: Okay I see. Do you need to also add the `global _pd_trace` to `infer_apply` as well then?

padl/transforms.py
sjrl: But in the case for `loader = list(enumerate(args))` the `ix` will be a single number right? This means running `[.. for i in ix]` will throw a `TypeError` because we would be iterating over an `int`.

padl/transforms.py
sjrl: Yeah I think it would be better to make a custom iterator to wrap around `args` to get it to do what you want.

padl/utils.py
```diff
@@ -61,3 +63,122 @@ def __getattr__(self, attr):
 
 #: Transform factory for capturing attributes/ get-items.
 same = _Same()
+
+
+class _Debug:
+    """Customized debugger for :class:`padl.Transform`s. When an exception on the execution of
+    a :class`Transform` is produced and a :class:`_Debug` object is called, an interactive debugger
```
sjrl, line 70: Missing `:` after `:class`

padl/utils.py
```diff
@@ -61,3 +63,122 @@ def __getattr__(self, attr):
...
+    def __init__(self):
+        self.trans = None
+        self.args = None
+        self.default_msg = (
+            'Defined commands are: \n'
+            '   u(p): step up\n'
+            '   d(own): step down\n'
+            '   w(here am I?): show code position\n'
+            '   i(nput): show input here\n'
+            '   r(epeat): repeat here (will produce the same exception)\n'
```
sjrl, line 89: I think we should consider removing the `r(epeat)` option since it currently requires us to store the original input arguments, which isn't ideal. Also looking into the `pdb` [debugger](https://www.uni-muenster.de/AMM/num/Vorlesungen/Pythonkurs_SS15/pdb_cheatsheet.pdf), I think this option is non-standard. What do you think @wuhu @eramdiaz ?

padl/transforms.py
```diff
@@ -861,10 +907,21 @@ def infer_apply(self, inputs):
...
+        _pd_trace.clear()
         inputs = self.pd_preprocess.pd_call_transform(inputs, mode='infer')
         inputs = _move_to_device(inputs, self.pd_device)
-        inputs = self.pd_forward.pd_call_transform(inputs, mode='infer')
-        inputs = self.pd_postprocess.pd_call_transform(inputs, mode='infer')
+        try:
+            inputs = self.pd_forward.pd_call_transform(inputs, mode='infer')
+        except Exception as err:
+            self._pd_trace_error(self._pd_get_error_idx('forward'), in_args)
+            raise err
```
sjrl, line 919: Why do we need to have the `try` and `except` to wrap the `pd_call_transform`, and also be present in the `_pd_call_transform`, which is called by `pd_call_transform`?

padl/transforms.py
eramdiaz: yes I think so

padl/transforms.py
eramdiaz: ah yes, I see, then what about writing a class like `_ItemGetter` for this case, but that retrieves a li. st of one item? For example: ([0], batch). However, the problem with storing the args on memory is still there isn't it?

padl/transforms.py
eramdiaz: This makes sense?

padl/transforms.py
eramdiaz: should I write the overloaded comparison now?

padl/transforms.py
eramdiaz: Do we know something?

padl/transforms.py
eramdiaz: Because in `pd_call_transform` we are tracing the error of `self.pd_preprocess`, `self.pd_forward` or `self.pd_postprocess`, which are the transforms that apply `pd_call_transform`. In this `try` we are adding one more element to the trace where we inspect the entire transform (`pd_preprocess`, `pd_forward` + `pd_postprocess`).

padl/utils.py
eramdiaz: I see it as a nice option to check again the error that occurred after inspecting inside the transform and making an hypothesis of the error. By the original input arguments, you refer to the absolute inputs to the transform, right? Keeping the arguments is also needed for the `i(nput)` option, which is very clarifying in my opinion.

padl/utils.py
eramdiaz: solved:)

padl/transforms.py
sjrl: Yes the memory issue with the args would remain.

padl/transforms.py
sjrl: It sounds like this could be solved by Issue #340 since resolving Issue #340 is awaiting the graph implementation of PADL.

padl/transforms.py
sjrl: @wuhu What do you think? Should we overload the comparison operator for transforms?

padl/transforms.py
sjrl: Not yet

padl/transforms.py
eramdiaz: I've made a new iterator similar to `_ItemGetter`. Actually, it would also be an option to introduce an argument in `_ItemGetter`to act in two different ways.

padl/transforms.py
eramdiaz: yes, so maybe we should remove this check, because it is also notworking on the 100% of cases

padl/utils.py
sjrl: > I see it as a nice option to check again the error that occurred after inspecting inside the transform and making an hypothesis of the error. 

I'm not sure I agree with this since I think you usually see the original error before entering a debugger right? 

> By the original input arguments, you refer to the absolute inputs to the transform, right? 

Yes

> Keeping the arguments is also needed for the `i(nput)` option, which is very clarifying in my opinion.

Yes I could see how this could be clarifying. However, isn't it possible to see what the inputs are through the traceback? For example, progressing upwards through the call tree up until the first call and printing the input variable?

padl/transforms.py
```diff
@@ -510,6 +516,24 @@ def f(x):
 
         return graph
 
+    def _pd_process_traceback(self):
+        """ Find where the Transform was defined (file, lineno, file) given the traceback. """
+        a_tb = None
+        for a_tb in self._pd_traceback[::-1]:
+            if 'padl' in a_tb[0]:
```
sjrl, line 523: What is this check specifically for? And why do we need to skip parts of the traceback with `padl` in it?

padl/transforms.py
```diff
@@ -1552,16 +1634,63 @@ def _pd_get_keys(transforms):
...
+    def _pd_get_non_target_stage_idx(self, stage_used: bool):
+        """When a :class:`Compose` gets an Exception and has a non-failing stage which is a
+        :class:`Parallel` or a :class:`Rollout`, the index they contribute to track the
+        element that got the Exception on the entire :class:`Compose` is 1.
+
+        Example:
+            t = ((prep_1 >> batch) + (prep_2 >> batch)) >> forward_1
+            Let's suppose we get an error on `forward_1`, so the item that is failing on `t` is 1.
+            `forward_1` is the index 0 of `t.pd_forward`, then the element that fails on `t` is
+            t._pd_preprocess._pd_get_non_target_stage_idx() + 0 = 1 + 0 = 1
```
sjrl, line 1646: `t._pd_preprocess` to `t.pd_preprocess`, it is not a hidden method.

padl/transforms.py
eramdiaz: Can we resolve this?

padl/transforms.py
```diff
@@ -1552,16 +1634,63 @@ def _pd_get_keys(transforms):
...
+        """When a :class:`Compose` gets an Exception and the failing stage is a :class:`Parallel`
+        or a :class:`Rollout`, the index they contribute with to track the element that got the
+        Exception on the entire :class:`Compose` is 0.
+
+        Example:
+            t = prep >> batch >> (forward_1 + forward_2)
+            Let's suppose we get an error on `(forward_1 + forward_2)`, so the item that is
+            failing on `t` is 2. `(forward_1 + forward_2)` is the index 0 of `t.pd_forward`,
+            then the element that fails on `t` is
+            2 + t._pd_forward._pd_get_stage_idx() = 2 + 0 = 2
```
sjrl, line 1662: `t._pd_forward` to `t.pd_forward`

padl/transforms.py
eramdiaz: I added it and all the tests are passing, so it should be ok

padl/transforms.py
```diff
@@ -1552,16 +1634,63 @@ def _pd_get_keys(transforms):
...
+
+        :param is_entire_transform: *True* if *self* constitutes entirely an overlying
+            :class:`Transform`, else *False*
```
sjrl, line 1665: I think this explanation is a little confusing because we have no definition of what an `overlying Transform` is. I think it would be clearer if you describe that `*self* is not the stage of a Pipeline` or something similar.

padl/transforms.py
sjrl: Yes I think so. However, in general I feel like the number of `try` and `except` clauses needed makes the code harder to read. I was hoping there would have been a way to reduce the number of `try`'s and `except`'s.

padl/transforms.py
eramdiaz: `traceback.extract_stack()` follows the lines of code for defining some object. The last element with `padl` on it contains the line where it's been defined. For example, if we create a `Transform` like
```
linear = transform(torch.nn.Linear)(6, 4)
```
The last element of the traceback with `padl` on the filename is:
```
'linear = transform(torch.nn.Linear)(6, 4)'
```

padl/transforms.py
eramdiaz: I understand, and this is something that happens on other parts of the docs, but these methods are applied to stages of a transform, and I don't see a way to explain what these methods do without referring to the overlying transform.

padl/transforms.py
eramdiaz: good catch

padl/transforms.py
eramdiaz: changing it

padl/utils.py
eramdiaz: mmm, honestly I don't know about that, maybe @wuhu has an idea about this. This may affect also to the in-line compiler at the debugger: at the moment we can give an standard input to the debugger like:
```
> i.shape
torch.Size([2, 3, 8]) # for example
```
I am not sure if this would work without storing the args

padl/transforms.py
sjrl: However, it looks like the tests are failing for python 3.7

padl/transforms.py
eramdiaz: Would it be possible to save `in_args` in disk with the `padl` value save in a temporary file?

padl/transforms.py
eramdiaz: Actually for the case of lists it is not needed to use the `global` keyword, because the `append` method adds an element on the memory associated to a list. This is different to what happens if you use a global `integer`, for example, and change its value inside the function.

For example:
```
a = 5

def something():
    a = 0
    print(f'a is {a}`)

something()
print(f' a is {a}')
```
prints:
```
a is 0
a is 5
```
but 
```
a = 5

def something():
    global a
    a = 0
    print(f'a is {a}`)

something()
print(f' a is {a}')
```
prints:
```
a is 0
a is 0
```
if the flobal variable is a list and inside `something` you append an element it is going to be added to the global variable no matter the `global` keyword` is called or not.

padl/transforms.py
wuhu: It would be possible, but you shouldn't do that as it would potentially create a lot of new problems (you would need to keep track of the files, find a good way to serialize any kind of arg etc).

As mentioned before, I don't think we have a problem here because nothing is being copied.

padl/utils.py
```diff
@@ -61,3 +63,122 @@ def __getattr__(self, attr):
...
+                msg = _pd_trace[pos].transform_str
+            else:
+                i = _pd_trace[pos].args
+                try:
+                    code = compile(x, '', 'single')
+                    exec(code)
+                except Exception as err:
+                    print(err)
+                finally:
+                    continue
```
wuhu, line 147: It seems like `continue` is not allowed inside of a `finally` block for python 3.7. I think you can just remove the `finally` block.

```
E     File "/home/runner/work/padl/padl/padl/utils.py", line 147
E       continue
E       ^i
E   SyntaxError: 'continue' not supported inside 'finally' clause
```

wuhu (changes requested):

padl/transforms.py
```diff
@@ -12,8 +12,11 @@
 from os import remove
 from shutil import rmtree
 import textwrap
+import traceback
 from tempfile import TemporaryDirectory
 import types
+from dataclasses import dataclass
+from typing import Any
```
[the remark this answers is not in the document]
eramdiaz: done

padl/transforms.py
```diff
@@ -510,6 +516,24 @@ def f(x):
 
         return graph
 
+    def _pd_process_traceback(self):
+        """ Find where the Transform was defined (file, lineno, file) given the traceback. """
```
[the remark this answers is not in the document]
eramdiaz: right

padl/transforms.py
```diff
@@ -510,6 +516,24 @@ def f(x):
...
+        a_tb = None
+        for a_tb in self._pd_traceback[::-1]:
+            if 'padl' in a_tb[0]:
+                continue
+            break
+        return f'{a_tb.filename} in {a_tb.name}\n----> {make_green(a_tb.lineno)}    {a_tb.line}'
+
+    def _pd_trace_error(self, position: int, arg):
+        """ Add some error description to `pd_trace`. """
```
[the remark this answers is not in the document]
eramdiaz: done:)

padl/transforms.py
eramdiaz: Yes, because `Pipeline`s and `AtomicTransform`s implement different `_pd_call_transform` method. This is because in `Pipeline`s the `try` and the Exception catching are done inside `_pd_call_transform`  and in `Pipeline`s this is inside the `__call__` method, so we can track the item `i` that is failing.

padl/transforms.py
eramdiaz: ok, then removed!

padl/transforms.py
eramdiaz: yes, the code has been changed. Before when `use_preprocess` was False, loader was just `list(enumerate(args))`, so in the statement
`for ix, batch in loader`, so `ix` in this case would be an integer. Then, the the line
```
self._pd_trace_error(self._pd_get_error_idx('forward'), [args[i] for i in ix])
```
would produce a `TypeError`.

padl/transforms.py
eramdiaz: done

padl/transforms.py
```diff
@@ -2184,24 +2425,78 @@ class _ItemGetter:
     :param default: The default value to fall back to in case of exception.
     """
 
-    def __init__(self, samples, transform, exception=None, default=None):
+    def __init__(self, samples, transform, entire_transform, exception=None, default=None):
```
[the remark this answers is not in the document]
eramdiaz: right

padl/transforms.py
```diff
@@ -2184,24 +2425,78 @@ class _ItemGetter:
...
-        return self.transform(self.samples[item])
+                return [item], self.default
+        return [item], self.samples[item]
 
     def __len__(self):
         return len(self.samples)
 
 
+@dataclass
+class _SingleTrace:
```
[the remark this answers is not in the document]
eramdiaz: ok, both look good, maybe it is better `_TraceItem` so we don't get confused with `_pd_trace`.

padl/utils.py
```diff
@@ -61,3 +63,122 @@ def __getattr__(self, attr):
 
 #: Transform factory for capturing attributes/ get-items.
 same = _Same()
+
+
+class _Debug:
+    """Customized debugger for :class:`padl.Transform`s. When an exception on the execution of
```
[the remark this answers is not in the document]
eramdiaz: done

padl/transforms.py
eramdiaz: For some reason creating the following method:
```
def __eq__(self, other):
    return str(self) == str(other)
```
is causing error in pytorch `TypeErrors` like:
```
TypeError: unhashable type: '<name_of_the_class>`
```
when excecuting statements like: 
```
memo = set()
if self not in memo:
    ....
```

padl/transforms.py
sjrl: The memory issue had to do with previously wrapping `args` with `list`, which could force the retrieval of all items in args if it happened to be an iterator. It's been resolved now by using `loader = _BatchGetter(args)`

padl/transforms.py
sjrl: You're right. It would only delay in garbage collecting of `_in_args`.

padl/utils.py
sjrl: Okay sounds good and I agree we are just delaying garbage collection.

padl/transforms.py
```diff
@@ -2184,24 +2425,78 @@ class _ItemGetter:
...
-    def __init__(self, samples, transform, exception=None, default=None):
+    def __init__(self, samples, transform, entire_transform, exception=None, default=None):
         self.samples = samples
         self.transform = transform
+        self.entire_transform = entire_transform
+        self.exception = exception
+        self.default = default
+
+    def __getitem__(self, item):
+        if self.exception:
```
sjrl, line 2436: Is it possible to rewrite this block like this?
```python
try:
    return item, self.transform(self.samples[item])
except self.exception:
    return item, self.default
except Exception as err:
    is_entire_transform = str(self.entire_transform) == str(self.entire_transform.pd_preprocess)
    self.entire_transform._pd_trace_error(
        self.entire_transform.pd_preprocess._pd_get_target_stage_idx(is_entire_transform),
        [self.samples[item]]
    )
    raise err
```

padl/transforms.py
```diff
@@ -2184,24 +2425,78 @@ class _ItemGetter:
...
+                self.entire_transform.pd_preprocess._pd_get_target_stage_idx(is_entire_transform),
+                [self.samples[item]]
+            )
+            raise err
+
+    def __len__(self):
+        return len(self.samples)
+
+
+class _BatchGetter:
```
sjrl, line 2456: @eramdiaz As you mentioned before we could probably use `_ItemGetter` instead of `_BatchGetter`. Instead of adding an option to `_ItemGetter` you could also pass the `Identity()` as the `transform` input argument for `_ItemGetter`.

padl/transforms.py
sjrl: It looks like it's possible to handle multiple exceptions in python as shown here https://stackoverflow.com/questions/6095717/python-one-try-multiple-except

padl/transforms.py
sjrl: Could you provide a minimal code example to reproduce the error?

eramdiaz: yes, adding the previous `__eq__` method to `Transform`, and doing, for example:

```
@transform
def to_tensor(x):
    return torch.LongTensor(x)

emb = transform(torch.nn.Embedding)(10, 8)

t = to_tensor >> batch >> emb

list(t.train_apply([[9, 8, 8], [4, 2, 2], [5, 9, 1], [3, 7, 6]], batch_size=2, num_workers=0))
```
or also running the tests after adding `__eq__`. I think it is getting in conflict with other `__eq__` methods.

eramdiaz: yes, adding the previous `__eq__` method to `Transform`, and doing, for example:

```
emb = transform(torch.nn.Embedding)(10, 8)
memo = set()
emb in memo
```
or also running the tests after adding `__eq__`. I think it is getting in conflict with other `__eq__` methods.

padl/transforms.py
eramdiaz: yes, but there is also one more thing that could overload the class, in `_BatchGetter` we are returning the index on a list, like 
`return [item], self.samples[item]`, while in `_ItemGetter` we are returning just the index `return item, self.samples[item]`

sjrl: In that case adding a simple check if `item` is a `list` or not would work, right?
```python
if isinstance(item, int):
     item = [item]
```

padl/transforms.py
```diff
@@ -1743,10 +1878,88 @@ def __call__(self, args):
...
+        preprocess_used = False if (isinstance(preprocess, Identity)) and \
+                                   not isinstance(self.__getitem__(0), Identity) else True
+
+        preprocess_idx = preprocess._pd_get_non_target_stage_idx(preprocess_used)
+        if stage == 'forward':
+            is_entire_transform = str(self) == str(forward)
+            return preprocess_idx + forward._pd_get_target_stage_idx(is_entire_transform)
+
+        forward_used = False if (isinstance(forward, Identity)) and not \
+            isinstance(self.__getitem__(preprocess_idx), Identity) else True
```
wuhu, line 1958: just ->
```
self[preprocess_idx]
```

padl/transforms.py
wuhu: quick solution to get around needing `__eq__`:
```
is_entire_transform = isinstance(self.pd_preprocess, Identity) and isinstance(self.pd_postprocess, Identity)
```

padl/transforms.py
eramdiaz: removed and added as part of issue #340

padl/transforms.py
```diff
@@ -1743,10 +1878,88 @@ def __call__(self, args):
...
+            Let's suppose we get an error on `forward_1`, so the item that is failing on `t` is 2.
+            `forward_1` is the index 0 of `t.pd_forward`, then the element that fails on `t` is
+            t._pd_preprocess._pd_get_non_target_stage_idx() + 0 = 2 + 0 = 1
+
+            prep = (prep_1 >> prep_2 >> batch) - 'prep_name'
+            t = prep >> forward_1
+            In this case, the index `t.pd_preprocess` contributes to the item that fails on `t` is
+            1.
+
+        :param stage_used: *True* if stage used on the :class:`Compose`, else *False*
```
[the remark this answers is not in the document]
eramdiaz: in the end we decided to remove that argument and include the case it was handling on the issue #340

padl/transforms.py
eramdiaz: The problem with that is that we are missing the `if self.exception` check, that allows us to make sure that `self.exception` is not `None` (the default value) and then we don't get the following error:
```
TypeError: catching classes that do not inherit from BaseException is not allowed
```
if we try to go into it.

padl/transforms.py
eramdiaz: In both cases `item` is an integer (because in the `use_preprocess` case we call later a `DataLoader` that makes the list of indices and the `args`). What could work using the current code would be call `_ItemGetter` with `Identity` and then `DataLoader` with `batch_size=1'.

eramdiaz: This solution is causing issues. It is changing datatypes because of internal `DataLoader` working mode and making a test fail.

padl/transforms.py
eramdiaz: Yes, I agree that the explanation is not very clear. I am trying to rephrase it. If I understood correctly, @sjrl suggested to describe this method without mentioning that `self` is an stage of a larger `Transform`. If I don't mention this relation, I don't see a way to explain accurately what's the meaning of this method or what we are doing inside.

eramdiaz: Yes, I agree that the explanation is not very clear. I am trying to rephrase it, but it is a bit difficult without mentioning that `self` is an stage of a larger `Compose` (this is what @sjrl suggested, if I understood correctly). I am upgrading to a more minimal (and I think clearer) explanation. Let me know how you like it.

sjrl: I think the current comment is incorrect.
```
:param is_entire_transform: *True* if *self* is a part of a larger :class:`Transform`,
            else *False*
```
I think you meant 
```
*True* if *self* is not part of a larger :class:`Transform`, else *False*
```

padl/transforms.py
wuhu, line 2441: please make it `if self.exception is not None:`

padl/transforms.py
eramdiaz: @wuhu @sjrl

padl/transforms.py
eramdiaz: @wuhu @sjrl

padl/transforms.py
```diff
@@ -723,20 +758,30 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
         if use_preprocess:
             loader = self.pd_get_loader(args, preprocess, mode, **loader_kwargs)
         else:
-            loader = args
+            loader = _SimpleGetter(args)
+#            loader = self.pd_get_loader(args, Identity(), mode, batch_size=1)
```
sjrl, line 762: Remove commented out line

padl/transforms.py
```diff
@@ -2191,28 +2433,84 @@ class _ItemGetter:
...
+
+class _SimpleGetter:
+    """A simple item getter.
+
+    :param samples: An object implementing __getitem__ and __len__.
+    :param exception: Exception to catch for (fall back to *default*).
+    :param default: The default value to fall back to in case of exception.
+    """
+
+    def __init__(self, samples, exception=None, default=None):
```
sjrl, line 2478: @eramdiaz I think we can remove the `exception` and `default` parameters here and simplify the `__getitem__` function. We only use `_SimpleGetter` internally and the `exception` and `default` parameters are never used. If we need them in the future we can add it back, but I can't think of a use case for them at the moment.

padl/utils.py
```diff
@@ -61,3 +63,129 @@ def __getattr__(self, attr):
...
+
+    def _repeat_entire(self, mode):
+        if mode == 'train':
+            list(self.trans.train_apply(self.args, batch_size=len(self.args), num_workers=0))
+        if mode == 'eval':
+            list(self.trans.eval_apply(self.args, batch_size=len(self.args), num_workers=0))
+        self.trans.infer_apply(self.args)
+
+    def _repeat_on_stage(self, mode):
+        self.trans.pd_call_transform(self.args, mode)
```
sjrl, line 188: Just to clarify, `self.trans` is really `pd_preprocess`, `pd_forward`, or `pd_postprocess` when this function is called right?

sjrl:
I think it would be nice if we could add a short explanation and example of how to use the Debugger in the sphinx docs. Maybe a new md doc under `padl/docs/src/usage/`?

padl/transforms.py
eramdiaz: I agree

eramdiaz: Do you agree @wuhu ?

padl/utils.py
eramdiaz: yes

padl/transforms.py
```diff
@@ -931,6 +986,36 @@ def _pd_direct_subtransforms(self) -> Iterator[Transform]:
...
+            Let's suppose we get an error on `forward_1`. `forward_1` is the index 0 of
+            `t.pd_forward`, then the element that fails on `t` is
+            2 + t.pd_forward._pd_get_stage_idx() = 2 + 0 = 2
+
+        :param is_entire_transform: *False* if *self* is not a part of a larger :class:`Transform`,
+            else *True*
+        """
+        return 0
+
+    def _pd_get_error_idx(self, stage: str):
```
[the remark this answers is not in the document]
eramdiaz: I thought it was clearer on `AtomicTransform` since different `Pipeline`s and `AtomicTransform` differ pretty much  and that method is only used on `AtomicTransform` and its subclasses, but I can change it.

padl/transforms.py
wuhu: agree!

padl/utils.py
```diff
@@ -61,3 +63,129 @@ def __getattr__(self, attr):
...
+            pos += 1
+            return pos, pd_trace[pos].transform_str
+        return pos, 'Reached top level.'
+
+    def _repeat_entire(self, mode):
+        if mode == 'train':
+            list(self.trans.train_apply(self.args, batch_size=len(self.args), num_workers=0))
+        if mode == 'eval':
+            list(self.trans.eval_apply(self.args, batch_size=len(self.args), num_workers=0))
+        self.trans.infer_apply(self.args)
```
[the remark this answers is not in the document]
wuhu: yes, like it is currently I think you're also always doing the infer apply for all the modes

padl/transforms.py
sjrl: Sounds good, having a separate Getter is a good solution then.

padl/transforms.py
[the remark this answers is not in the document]
sjrl: Right, I see. In that case you could also provide a `NotImplemented` version of the method in `Transform` class to make it clear to developers in the future that the method needs to be implemented for new internal Transforms that might not inherit from `AtomicTransform` or `Pipeline`

padl/utils.py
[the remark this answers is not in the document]
eramdiaz: > yes, like it is currently I think you're also always doing the infer apply for all the modes

What do you mean? I did not get that completely @wuhu

padl/transforms.py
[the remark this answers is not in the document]
eramdiaz: ah sure, that sounds great

padl/utils.py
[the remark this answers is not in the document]
wuhu: I guess it's outdated now, I just meant that before there were no returns in the if blocks. But I didn't think about the exception, so never mind.

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+t = to_tensor >> batch >> emb >> linear
+```
+
+Then, if we run
+
+```python
+list(t.train_apply([[9,8,8], [4, 4, 4], [5, 5, 5], [6, 6, 6]], batch_size=2, num_workers=0))
+```
+
+we get an error since the shape of the embeddings does not match wit the one expected by the linear
```
sjrl, line 51: typo `wit` to `with`

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+{class}`~padl.transforms.Pipeline` and inspect relevant information on each level.
+
+The debugger console has defined the following commands:
+```
+* u(p): go one abstraction level up on the {class}`~padl.transforms.Pipeline`.
+* d(own): go one abstraction level down on the {class}`~padl.transforms.Pipeline`.
+* w(here am I?): show code position where the {class}`~padl.transforms.Transform` was defined.
+* i(nput): show input to this level.
+* h(elp): print help about the commands.
+* q(uit): quit the debugger.
```
sjrl, line 25: I think there are a few options missing like `r(epeat)`, right?

padl/transforms.py
```diff
@@ -146,6 +151,8 @@ def __init__(self, call_info: Optional[inspector.CallInfo] = None,
         self._pd_varname = {}
         self._pd_name = pd_name
         self._pd_device = 'cpu'
+        self._pd_traceback = traceback.extract_stack()
+        self._pd_stages = None
```
sjrl, line 155: I don't think `self._pd_stages` is used anymore. It may be a left over after merging in main. Please remove

padl/transforms.py
```diff
@@ -508,6 +515,27 @@ def f(x):
...
+    def _pd_process_traceback(self):
+        """Find where the Transform was defined (file, lineno, file) given the traceback. """
+        a_tb = None
+        for a_tb in self._pd_traceback[::-1]:
+            if 'padl' in a_tb[0]:
+                continue
+            break
+        return f'{a_tb.filename} in {a_tb.name}\n----> {make_green(a_tb.lineno)}    {a_tb.line}'
+
+    def _pd_get_error_idx(self):
```
sjrl, line 527: Add docstring

padl/transforms.py
```diff
@@ -882,11 +936,19 @@ def infer_apply(self, inputs):
         if use_preprocess:
             inputs = preprocess.pd_call_in_mode(inputs, mode='infer', ignore_grad=True)
         if pd_device != 'cpu':
-            inputs = _move_to_device(inputs, pd_device)
+            inputs = _move_to_device(inputs, self.pd_device)
```
sjrl, line 939: Change `self.pd_device` to local variable `pd_device` since it's already been defined.

padl/utils.py
```diff
@@ -61,3 +63,131 @@ def __getattr__(self, attr):
 
 #: Transform factory for capturing attributes/ get-items.
 same = _Same()
+
+
+class _Debug:
+    """Customized debugger for :class:`padl.Transform`s.
+
+    When an exception on the execution of a :class:`Transform` is produced and a :class:`_Debug`
```
sjrl, line 71: Be consistent with reference to `Transform`
```python
:class:`Transform`
```
or 
```python
:class:`padl.Transform`
```

tests/unittests/test_transforms.py
```diff
@@ -959,3 +959,26 @@ def test_double_4(self):
...
+    def init(self, request):
+        emb = transform(torch.nn.Embedding)(10, 8)
+        linear = transform(torch.nn.Linear)(8, 4)
+        to_tensor = transform(lambda x: torch.LongTensor(x))
+        request.cls.pipeline = to_tensor >> emb >> linear
+
+    def test_pd_trace(self):
+        try:
+            list(self.pipeline.train_apply([[9, 8, 8], [4, 4, 4], [5, 5, 5], [6, 6, 6]],
+                                           batch_size=2, num_workers=0))
```
sjrl, line 975: It doesn't look like this `pipeline` produces an `RuntimeError` exception. I think the linear layer above needs to be changed to something like `linear = transform(torch.nn.Linear)(5, 4)`

tests/unittests/test_transforms.py
```diff
@@ -959,3 +959,26 @@ def test_double_4(self):
...
+        except:
+            from padl.transforms import _pd_trace
+            len(_pd_trace) == 3
```
sjrl, line 978: Please change add `assert` to each of these statements otherwise `pytest` can't check if they all evaluate to `True`

docs/src/usage/debugging_your_transforms.md
eramdiaz: yes, i missed them

tests/unittests/test_transforms.py
eramdiaz: ups sure, thanks

docs/src/usage.md
```diff
@@ -35,3 +35,4 @@ usage/extras.md
 - {class}`~padl.IfEval` for conditioning on the 'eval'-stage.
 - {class}`~padl.IfTrain` for conditioning on the 'train'-stage.
 - {class}`~padl.Try` for catching exceptions in a transform.
+- {class}`~padl.utils.pd_debug` for debugging Transforms.
```
sjrl, line 38: In this file you also need to add `usage/debugging_your_transforms.md` to the `{toctree}` so your `.md` file is added to the `html`

tests/unittests/test_transforms.py
eramdiaz: right, missed it

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+layer. Calling {obj}`~pd_debug` after this will retrieve a console where we can make some checks. 
+
+At the starting level of the debugger, we interact with the entire Transform and its absolute 
+input. One level down, it goes directly to the **stage** that got the Exception (*preprocess*, 
+*forward* or *postprocess*), and each level deeper moves recursively inside the element that failed
+until the {class}`~padl.transforms.AtomicTransform` that failed is reached.
+
+Below, some examples are provided:
+
+```python
```
sjrl, line 61: Before you start using debugging commands I would show the code to import and call `pd_debug` to make it clear in the code that this needs to be done. So something like 
```python
list(t.train_apply([[9,8,8], [4, 4, 4], [5, 5, 5], [6, 6, 6]], batch_size=2, num_workers=0))

pd_debug()
```

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+   0: to_tensor                                    
+      │
+      ▼ args
+   1: Batchify(dim=0)                              
+      │
+      ▼ input
+   2: Embedding(num_embeddings=10, embedding_dim=8)
+      │
+      ▼ input
+   3: Linear(-?-)                                    <---- error here
```
sjrl, line 81: Also out of curiosity why does `Linear` have `-?-` but `Embedding` shows the correct variable calls?

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+
+torch.Size([2, 3, 8])
+
+> a = 4
+> a
+
+4
+
+> q
+```
```
sjrl, line 110: Overall this example is very nice! My only suggestion would be to show more of the other functionality options like `w` and `r`.

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+* w(here am I?): show code position where the {class}`~padl.transforms.Transform` was defined.
+* i(nput): show input to this level.
+* h(elp): print help about the commands.
+* q(uit): quit the debugger.
+```
+
+On top of this, it incorporates an in-line compiler so the user can define and inspect variables, 
+check attributes, import packages...
+
+#### Example
```
wuhu, line 31: Use ### here as it's one level below.

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Then, if we run
+
+```python
+list(t.train_apply([[9,8,8], [4, 4, 4], [5, 5, 5], [6, 6, 6]], batch_size=2, num_workers=0))
+```
+
+we get an error since the shape of the embeddings does not match wit the one expected by the linear
+layer. Calling {obj}`~pd_debug` after this will retrieve a console where we can make some checks.
```
wuhu, line 52: add `padl` (see above)

docs/src/usage/debugging_your_transforms.md
wuhu: there seems to be a bug, I'd suggest creating a ticket and change it here now so it looks it would if it were working

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
...
+
+```python
+list(t.train_apply([[9,8,8], [4, 4, 4], [5, 5, 5], [6, 6, 6]], batch_size=2, num_workers=0))
+```
+
+we get an error since the shape of the embeddings does not match wit the one expected by the linear
+layer. Calling {obj}`~pd_debug` after this will retrieve a console where we can make some checks. 
+
+At the starting level of the debugger, we interact with the entire Transform and its absolute 
+input. One level down, it goes directly to the **stage** that got the Exception (*preprocess*,
```
wuhu, line 55: how about reminding people how to go down, something like

"When going down one level by entering "d", the debugger switches to..."

docs/src/usage/debugging_your_transforms.md
```diff
@@ -0,0 +1,111 @@
+(debugging-your-transforms)=
+## Debugging your Transforms
+
+PADL provides a customized debugger to inspect and diagnose errors raised when applying a 
+{class}`~padl.transforms.Pipeline`.
+
+For using this functionality, call {obj}`~pd_debug` after the execution failed, such as
```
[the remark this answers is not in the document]
eramdiaz: Then would you create it or would you leave it like `pd_debug`?

docs/src/usage/debugging_your_transforms.md
wuhu: I think it could be useful to link to the "stages" chapter here:
```
{ref}`stage <stages>`
```
instead of
```
**stage**
```

padl/transforms.py
```diff
@@ -1655,12 +1802,19 @@ def _pd_splits(self, input_components=0, has_batchify=False) -> Tuple[Union[int,
...
-                final_splits.append(split[0])
+                if isinstance(split[0], Compose):
+                    final_splits.append(group(split[0]))
+                else:
+                    final_splits.append(split[0])
             else:  # if it's empty: identity
                 final_splits.append(identity)
 
-        self._add_name_to_splits(final_splits)
+        if self._pd_name is not None:
```
wuhu, line 1812: you can remove this as it's done already in `_add_name_to_splits`

padl/transforms.py
```diff
@@ -1883,8 +2104,12 @@ def _pd_splits(self, input_components=0) -> Tuple[Union[int, List],
                 res.append(split)
         final_splits = res
 
-        self._add_name_to_splits(final_splits)
+        if self._pd_name is not None:
```
wuhu, line 2107: you can remove this as it's done already in _add_name_to_splits

docs/src/usage/debugging_your_transforms.md
eramdiaz: Thanks! I am adding `w`. The thing about `r` is that the traceback is very long. I can add some `...` and display only the last line (where the `RuntimeError` is described) but then I wouldn't use `q` because `r` gets out of the debugger by itself. Is it ok then?

padl/transforms.py
```diff
@@ -1978,8 +2210,11 @@ def _pd_splits(self, input_components=0) -> Tuple[Union[int, List],
                 res.append(split)
         final_splits = res
 
+        if self._pd_name is not None:
```
wuhu, line 2213: you can remove this as it's done already in _add_name_to_splits

docs/src/usage/debugging_your_transforms.md
sjrl: Ahh I see, then you can leave `r` out then. Otherwise I think it looks good.

docs/src/usage/debugging_your_transforms.md
eramdiaz: I was considering this but i did not want to mix the first reference or its presentation with an `apply` in the beggining. It is explained in this line
```
For using this functionality, call `{obj}`~pd_debug`` after the execution failed
```
which is followed by
```
from padl import pd_debug

pd_debug()
```

But I see your point, what do you think about writing it like
```
# train_apply, eval_apply or infer_apply that produced a RuntimeError

from padl import pd_debug

pd_debug()
```

docs/src/usage/debugging_your_transforms.md
eramdiaz: ok :)

padl/utils.py
```diff
@@ -61,3 +63,131 @@ def __getattr__(self, attr):
...
+    def repeat(self, mode: str, pos: int) -> None:
+        """Repeat the execution from the current position *pos* (the same Exception will be
+        produced).
+
+        :param mode: mode ('train', 'eval', 'infer').
+        :param pos: level of the :class:`Transform` we are inspecting.
+        """
+        assert mode in ('train', 'eval', 'infer'), 'Mode should be "train", "eval" or "infer'
+        if pos == len(_pd_trace) - 1:
+            _pd_trace.clear()
```
wuhu, line 161: you can move the `_pd_trace.clear()` before the `if`

docs/src/usage/debugging_your_transforms.md
sjrl: > But I see your point, what do you think about writing it like
> 
> ```
> # train_apply, eval_apply or infer_apply that produced a RuntimeError
> 
> from padl import pd_debug
> 
> pd_debug()
> ```

I think this would be good to have.

padl/utils.py
wuhu: Even though it won't get to the `_repeat_on_stage` if executing `_repeat_entire` due to the exception, for me it would look cleaner if you wrap `_repeat_on_stage` in `else`
```
if pos = ..:
    self._repeat_entire(mode)
else:
   self._repeat_on_stage(mode)
```

wuhu (changes requested):
Looks good!
Just a few small things remain.

docs/src/usage/debugging_your_transforms.md
eramdiaz: nice

padl/transforms.py
```diff
@@ -508,6 +514,32 @@ def f(x):
...
+            if 'padl' in a_tb[0]:
+                continue
+            break
+        return f'{a_tb.filename} in {a_tb.name}\n----> {make_green(a_tb.lineno)}    {a_tb.line}'
+
+    def _pd_get_error_idx(self):
+        """Get what element of a {class}`Transform` is failing if an Exception is produced during
+        an execution.
+
+        Subclasses of {class}`padl.transforms.Transform` need to implement this method.
```
sjrl, line 530: Thanks for the docstring! I think you accidentally used the wrong formatting for `{class}` in two places. I think it should be `:class:`.

eramdiaz: good catch

docs/src/usage/debugging_your_transforms.md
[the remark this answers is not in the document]
sjrl: Yes I think change it to 
```python
{obj}`~padl.pd_debug`
```

padl/utils.py
```diff
@@ -61,3 +63,131 @@ def __getattr__(self, attr):
...
+            '   q(uit): quit'
+            '     -> this will store the input at this level in debug.args\n'
+            '        and the transform in debug.trans\n'
+        )
+
+    def __call__(self) -> None:
+        """Call me for getting an interactive debugger in case of error.
+
+        User can give following input and expect response
+            u(p): step up\n'
```
sjrl, line 103: I don't think you need the newline characters here since you have used a triple quote to wrap the doc string.

padl/transforms.py
```diff
@@ -508,6 +514,32 @@ def f(x):
...
+        return NotImplemented
+
+    def _pd_trace_error(self, position: int, arg):
+        """Add some error description to :obj:`pd_trace`. """
+        try:
+            str_ = self._pd_fullrepr(marker=(position, '\033[31m  <---- error here \033[0m'))
+            _pd_trace.append(_TraceItem(str_, self._pd_process_traceback(), arg,
+                                        self, Transform.pd_mode, position))
+        except Exception:
+            warn('Error tracing failed')
```
sjrl, line 541: Sorry one additional question. Would it be worthwhile to raise the error here in the case the trace does fail?

padl/utils.py
```diff
@@ -61,3 +63,131 @@ def __getattr__(self, attr):
...
+            '   u(p): step up\n'
+            '   d(own): step down\n'
+            '   w(here am I?): show code position\n'
+            '   i(nput): show input here\n'
+            '   r(epeat): repeat here (will produce the same exception)\n'
+            '   t(ransform): displays the current transform\n'
+            '   h(elp): print help about the commands\n'
+            '   q(uit): quit'
+            '     -> this will store the input at this level in debug.args\n'
+            '        and the transform in debug.trans\n'
```
sjrl, line 96: Change `debug.args` to `pd_debug.args` and `debug.trans` to `pd_debug.trans`. Also was `-> this will ...` intended to be on a newline?

sjrl:
I also found an edge case when playing around with the debugger. 
```python
from padl import transform, pd_debug
t = transform(lambda x: x)
t2 = transform(lamba x: x + 's')

t2.infer_apply(2)
# Returns a TypeError

pd_debug() # works great! 

t.infer_apply(x) # x is purposely not defined
# Returns a NameError since x is not defined

pd_debug() # Still points to the trace from t2.infer_apply(2)
debug # debugger from ipython points to the trace from t.infer_apply(x)
```

sjrl:
I found an unexpected behavior (I think).
```python
t2.infer_apply(2)
pd_debug()
> x # This prints "x" instead of returning a "name x is not defined"
> y # This prints the expected "name y is not defined"
```
I'm not entirely sure why this happens. Do you know @eramdiaz ?

padl/transforms.py
eramdiaz: if we execute this block it means that we've got an error on a transform execution, and if we raise an error here, we would show two errors in the traceback: the runtime error of the transform and the one getting _pd_trace. This could in my opinion hide a bit the main error, which is the RuntimeError got while applying.

padl/utils.py
eramdiaz: I am reconsidering if `-> this will ...` is helpful anymore

padl/transforms.py
sjrl: Ahh I see now. I thought this would swallow the error from the transform execution, but that is not the case. Okay this looks good!

sjrl:
> I see, in the code there are a couple of `_pd_trace.clear()` statements to prevent not cleaning previous traces in new debugging sessions. In this case you are describing, the `pd_trace.clear()` inside `infer_apply` does not work because if `x` is not defined, then the code inside `infer_apply` is not even run (I would say that we are not even reaching the `padl` code), so I don't know how to solve it. In other case where we get into the `padl` code (for example, if you run again something like `t2.infer_apply(3)' , the trace is cleaned and it works as expected.
> 
> Otherwise, I'd recommend to use `pd_debug` with `Pipeline`s and avoiding using it with `AtomicTransform`s

Okay I see what you're saying. I think it's fine to leave as is.

sjrl (approved):

docs/src/usage/debugging_your_transforms.md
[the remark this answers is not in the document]
wuhu: the `~` makes it not display the `padl`

wuhu (approved):
Looks good!

sjrl:
@eramdiaz Please merge when you can

PR #341, opened 2021-12-24
Speedup

wuhu:
- add caching
- remove contexts
- don't set mode in subtransforms
- rename `pd_call_transform` to `pd_call_in_mode`

### DIFF
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -757,12 +757,12 @@ class Transform:
 
                 output = batch
                 if use_forward:
-                    output = forward.pd_call_transform(batch, mode)
+                    output = forward.pd_call_in_mode(batch, mode)
 
                 if use_post or flatten:
                     output = _unpack_batch(output)
                     if use_post:
-                        output = [post.pd_call_transform(x, mode) for x in output]
+                        output = [post.pd_call_in_mode(x, mode) for x in output]
                     for out in output:
                         yield self._pd_format_output(out)
                 else:
@@ -871,7 +871,7 @@ class Transform:
         """
         sequence = _ItemGetter(
             args,
-            lambda *args: preprocess.pd_call_transform(*args, mode),
+            lambda *args: preprocess.pd_call_in_mode(*args, mode, ignore_grad=True),
         )
 
         return DataLoader(
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -160,7 +160,7 @@ class Transform:
         a different module. Else, only dump an import statement. """
         module = self._pd_full_dump_relevant_module
         # always fully dump Transforms from the module the dump was triggered in
-        if inspector.caller_module() == module:
+        if inspector.caller_module() == module or getattr(module, '__name__', '__main__') == '__main__':
             return True
         # fully dump all Transforms from packages or modules specified in
         # _pd_external_full_dump_modules
@@ -1097,7 +1097,7 @@ class ClassTransform(AtomicTransform):
         instance_scope = self._pd_call_info.scope
 
         # instance creation and class definition are in separate modules
-        if instance_scope.module.__name__ != self.__class__.__module__:
+        if instance_scope.module_name != self.__class__.__module__:
             return super()._pd_codegraph_add_startnodes(graph, name)
 
         # the instance has a varname - just import the instance
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -888,10 +888,14 @@ class Transform:
         :param inputs: The input.
         """
         self.pd_forward_device_check()
-        inputs = self.pd_preprocess.pd_call_transform(inputs, mode='infer')
-        inputs = _move_to_device(inputs, self.pd_device)
-        inputs = self.pd_forward.pd_call_transform(inputs, mode='infer')
-        inputs = self.pd_postprocess.pd_call_transform(inputs, mode='infer')
+        if not isinstance(self.pd_preprocess, Identity):
+            inputs = self.pd_preprocess.pd_call_in_mode(inputs, mode='infer')
+        if self.pd_device != 'cpu':
+            inputs = _move_to_device(inputs, self.pd_device)
+        if not isinstance(self.pd_forward, Identity):
+            inputs = self.pd_forward.pd_call_in_mode(inputs, mode='infer')
+        if not isinstance(self.pd_postprocess, Identity):
+            inputs = self.pd_postprocess.pd_call_in_mode(inputs, mode='infer')
         return self._pd_format_output(inputs)
 
     def eval_apply(self, inputs: Iterable, flatten: bool = False, **kwargs):
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -684,7 +684,7 @@ class Transform:
 
         :param arg: argument to call the transform with
         :param mode: The mode ("infer", "eval", "train") to perform the call with.
-        :param ignore_grad: If *True* gradients are disabled
+        :param ignore_grad: If *True* gradient settings are ignored
         :return: Whatever the transform returns.
         """
         no_grad = mode in ('eval', 'infer') and not ignore_grad
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -190,6 +190,7 @@ class Transform:
             return Transform._pd_merge_components(res)
         return res
 
+    @property
     @lru_cache(maxsize=128)
     def _pd_get_stages(self):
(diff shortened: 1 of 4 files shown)


padl/transforms.py
```diff
@@ -672,23 +674,47 @@ def _pd_unpack_argument(self, arg) -> bool:
             return True
         return False
 
-    def pd_call_transform(self, arg, mode: Optional[Mode] = None):
+    def _pd_unpack_args_and_call(self, arg):
+        if self._pd_unpack_argument(arg):
+            return self(*arg)
+        return self(arg)
+
+    def pd_call_in_mode(self, arg, mode: Mode, ignore_grad=False):
```
sjrl, line 682: add doc string for `ignore_grad`

sjrl:
Nicely done!

sjrl:
@wuhu By the way, the solution for python 3.7 failing is to change the `@lru_cache` decorator to
```python
@functools.lru_cache(maxsize=None)
```
Found solution here: https://github.com/izar/pytm/issues/114

padl/transforms.py
```diff
@@ -758,80 +788,52 @@ def pd_device(self) -> str:
...
+        pre.pd_set_device(self.pd_device)
         return pre
 
     @property
     def pd_forward(self) -> "Transform":
         """The forward part of the transform (that what's typically done on the GPU).
         The device must be propagated from self."""
         forward = self._pd_get_stages()[1]
-        forward.pd_to(self.pd_device)
+        forward.pd_set_device(self.pd_device)
```
sjrl, line 799: @wuhu I replaced `pd_to` here with `pd_set_device` because that was the original intention of adding `pd_to` here. It was to make sure that the splits inherited the correct device from the parent transform. However, sending the layers to device multiple times creates unnecessary overhead hence the change.

wuhu: :+1:

padl/transforms.py
```diff
@@ -749,6 +776,9 @@ def _gen():
 
         return _GeneratorWithLength(_gen, length)
 
+    def pd_set_device(self, device: str):
```
wuhu, line 779: @sjrl I'd suggest making this a setter for the `pd_device` property (`@pd_device.setter`)

sjrl:
> @sjrl should we merge?

One minute. I added some accidental changes to one of the notebooks

sjrl:
I think it's good to go. I think we have gotten all of the easy speed ups for the `pd_call_in_mode` function. Issue #342 still has two more proposals which we haven't implemented yet related to speeding up `infer_apply` (removal of forward_device_check) and putting `_move_to_device` back into Batchify.

padl/transforms.py
```diff
@@ -662,33 +667,64 @@ def _pd_unpack_argument(self, arg) -> bool:
...
         """
+        no_grad = mode in ('eval', 'infer') and not ignore_grad
 
-        if mode in ('eval', 'infer'):
-            torch_context = torch.no_grad()
-        else:
-            torch_context = contextlib.suppress()
+        layers = self.pd_layers
+        if not layers:
+            no_grad = False
```
sjrl, line 698: @wuhu I added this check for the situation where `pd_forward` is not Identity but does not contain any torch layers.

padl/transforms.py
```diff
@@ -658,38 +663,71 @@ def _pd_unpack_argument(self, arg) -> bool:
...
+                        layer = layers[i]
+                        if training:
+                            layer.train()
+                        else:
+                            layer.eval()
+            if no_grad:
+                torch.set_grad_enabled(grad_before)
 
+    @property
+    @lru_cache(maxsize=128)
```
sjrl, line 730: @wuhu I changed this to be a property and cached since I believe it only needs to be evaluated once. Do you think this is correct?

padl/transforms.py
wuhu: :+1:

padl/transforms.py
wuhu: Agree to having the cache. If we make it a property we should rename it (to `_pd_signature`).

sjrl: Sounds good, I'll change this

padl/transforms.py
```diff
@@ -190,6 +190,8 @@ def _pd_merge_components(components):
             return Transform._pd_merge_components(res)
         return res
 
+    @property
+    @lru_cache(maxsize=128)
```
sjrl, line 194: @wuhu Since this is cached now do we need to store it into `self._pd_stages`? Could we do something like this 
```python
    @property
    @lru_cache(maxsize=128)
    def _pd_stages(self):
        _, splits, has_batchify = self._pd_splits()
        if has_batchify:
            preprocess, forward, postprocess = splits
        else:
            preprocess, forward, postprocess = identity, splits[0], splits[2]
        return preprocess, forward, postprocess
```

wuhu: yes! makes sense

padl/transforms.py
```diff
@@ -754,84 +793,60 @@ def pd_device(self) -> str:
...
     def pd_layers(self) -> List[torch.nn.Module]:
         """Get a list with all pytorch layers in the transform (including layers in sub-transforms).
         """
-        layers = []
-        for subtrans in self._pd_all_transforms():
-            if isinstance(subtrans, torch.nn.Module):
-                layers.append(subtrans)
-        return layers
+        if self._pd_layers is None:
```
wuhu, line 837: as this is being cached now, do we need to do this?

padl/transforms.py
```diff
@@ -658,38 +663,71 @@ def _pd_unpack_argument(self, arg) -> bool:
             return self._pd_number_of_inputs > 1
 
         try:
-            parameters = self._pd_get_signature().values()
+            parameters = self._pd_get_signature.values()
         except ValueError:
             return False
         for param in parameters:
-            if param.kind in (
+            param_kind = param.kind
```
wuhu, line 670: why this?

padl/transforms.py
```diff
@@ -639,12 +641,15 @@ def pd_forward_device_check(self) -> bool:
         All transforms in forward need to be in same device as specified for
         the whole Pipeline.
         """
-        for layer in self.pd_forward.pd_layers:
+        pd_device = self.pd_device
+        pd_layers = self.pd_forward.pd_layers
```
wuhu, line 645: I don't think that's necessary, `self.pd_device` is being cached and pd_layers is only used once below.

padl/transforms.py
```diff
@@ -639,12 +641,15 @@ def pd_forward_device_check(self) -> bool:
...
+        for layer in pd_layers:
             for parameters in layer.parameters():
                 parameter_device = parameters.device.type
-                if ':' in self.pd_device and 'cuda' in parameter_device:
-                    parameter_device += f':{parameters.device.index}'
-                if parameter_device != self.pd_device:
+                parameter_device_index = parameters.device.index
```
wuhu, line 649: Same here.

padl/transforms.py
sjrl: No we do not, I'll change this as well.

padl/transforms.py
sjrl: I don't think `self.pd_device` is being cached.

padl/transforms.py
sjrl: changing back

padl/transforms.py
sjrl: I agree with self.pd_forward_layers

padl/transforms.py
sjrl: Small optimization where an attribute lookup `param.kind` is slower than accessing a local variable (e.g. `param_kind`). Since we use `param.kind` more than once I assigned it to a local variable especially since this function is called a lot when evaluating a transform.

sjrl:
@wuhu I think this is ready to go. Do you agree?

padl/transforms.py
wuhu: Eh sorry not cached, but it's just a simple lookup, right? and this isn't called too often, so I'm not sure this warrants the slightly diminished readability. But either way is fine.

padl/transforms.py
wuhu: You resolved, but it's still the same isn't it? Point was that as parameter_device_index is only used once we don't need to assign it to a variable.

sjrl: I believe I changed this back. The code snippet is shown to be out of date at the top of this comment.

PR #348, opened 2022-01-13
Issue 346

### LINKED ISSUE #346: Fix image and version links in docs

sjrl:
## Description

Fixes #346


.github/workflows/build_docs.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -e .
```
sjrl, line 30: @wuhu This was the bug preventing the apidocs from building correctly. We go onto to change directory to `padl/docs`, and the apidocs when building expects to be able to `import padl`

docs/Makefile
```diff
@@ -27,6 +27,7 @@ build:
 	python gen_doc_files.py
 	make html
 	mv _build/html unstable
+	cp -r src/img unstable/
```
sjrl, line 30: @wuhu This was missing so the images would be properly linked. Out of curiosity, would it be "more" correct to store the images in the `static` folder and link to them in the `.md` files using `_static/my_image.png`?

docs/src/_templates/versions.html
```diff
@@ -15,17 +15,18 @@
     </div>
     <script>
         console.log('script')
+        var pathname = window.location.pathname.split('padl/')[1].split('/');
+        var currentVersion = pathname[1];
+        var versionsFile = window.location.pathname.split('padl/')[0] + 'padl/docs/versions.json';
         httpRequest = new XMLHttpRequest();
         httpRequest.onreadystatechange = addVersions;
-        httpRequest.open('GET', '../versions.json');
+        httpRequest.open('GET', versionsFile);
```
sjrl, line 23: @wuhu I'm not super confident with javascript so I'd appreciate if you could take a look at the changes I made here. I made them with the intent that the `versions.json` file will always be found no matter how far down you progress in the html directories.

wuhu (approved):

PR #365, opened 2022-01-25
Use pd_to in the pd_pre, pd_for, and pd_post properties

sjrl:
## Description

This is reverting a change made in PR #341 . `pd_to` needs to be used when creating `pd_preprocess`, `pd_forward`, and `pd_postprocess` otherwise the children transforms will not correctly inherit the correct device.


sjrl:
I noticed this when running the Sentiment Analysis notebook on Google Colab

PR #384, opened 2022-02-07
Linting

wuhu:
Some linting of `transforms.py`.


padl/transforms.py
```diff
@@ -806,7 +808,7 @@ def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
             self.pd_forward_device_check()
 
         if use_preprocess:
-            loader = self.pd_get_loader(args, preprocess, mode, **loader_kwargs)
+            loader = self._pd_get_loader(args, preprocess, mode, **loader_kwargs)
```
sjrl, line 811: I think we made this public because it will be called explicitly in the Pytorch Lightning connector. Should be fine either way though.

sjrl (approved):
Looks good!

sjrl:
Also could you update the documentation for Function Transform to replace `inline_wrap`?

padl/transforms.py
wuhu: ok I'll change it back

sjrl (approved):

Issue #340, opened 2021-12-21
Debugger not tracking the error properly when trivial Identities


eramdiaz:
## 🐞 Bug

When we get an exception on a transform that has useless `Identity()`s, the debugger is not able to track correctly which element of the entire `transform` is getting the Exception. 

Example: 
Let  `t` be the following Transform
```
t = prep_1 >> Identity() >> Identity() >> batch >> Identity() >> emb >> linear
```
and suppose we are getting an Exception on `linear`. `pd_splits` does not take into account the useless Identities, so `t.pd_preprocess`is
```
prep_1 >> batch
```
and `t.pd_forward` is 
```
emb >> linear
```
so when we catch the Exception on `t.pd_forward`, the item we get failing is `1`, and when we are trying to track this index back in the entire transform we get: 
```
len(t.pd_preprocess)` + failing_item = 2 + 1 = 3
```
so we would get that the transform that is faling is `batch`, which is wrong.

For solving this, we would need to count how many useless `Identities` are before the index failing, and I have not found a solution with the information we've got (the index of the exception on the stage that fails).

sjrl:
@wuhu Would one solution to this be to never remove the useless Identities.

wuhu:
@sjrl I think so. We would end up with a lot of identities in the splits though (this is why we remove them), possibly we could somehow not remove them when they were in there in the first place (in the original transform). 
Anyhow, I think this is a rather edgy edgecase and I think we'll probably be able to solve this easier once we move to the graph model. Therefore, I suggest leaving it as it is for now.

eramdiaz:
This issue was tackled on the pr https://github.com/lf1-io/padl/pull/405, where the first arrow of the debugger was removed due to this issue (https://github.com/lf1-io/padl/issues/340) and on https://github.com/lf1-io/padl/issues/404. At the moment we'll keep it open to take into account on a future implementation of this arrow.

PR #405, opened 2022-02-22
Remove first marker

### LINKED ISSUE #404: Debugger marker arrow not holding for composes with parallel or rollout preprocesses
## 🐞 Bug

There is a bug on the marker arrow at the highest level of the debugger when we are handling `Compose`s whose preprocess part is a `Parallel` or a `Rollout`.

When executing the debugger, the first level shows the entire transform we are applying and an arrow pointing to the step that is caused the exception. 

For example, let `t0` be the transform:
```
t0 = (transform(lambda x: torch.LongTensor(x))
    >> batch
    >> transform(torch.nn.Embedding)(10, 8)
    >> transform(torch.nn.Linear(6, 4)
)
```
When executing `t0`, an error is going to occur at the step 3 since the shapes are not matching. Then, if we call the debugger, its first level is going to show
```
 Compose - "t0":

      │
      ▼ x
   0: lambda x: torch.LongTensor(x)                
      │
      ▼ args
   1: Batchify(dim=0)                              
      │
      ▼ input
   2: Embedding(num_embeddings=10, embedding_dim=8)
      │
      ▼ input
   3: Linear(in_features=6, out_features=4)          <---- error here 
```
The way we determine where to put the arrow `<---- error here` in `Compose`s is done (when the `Compose` fails in the forward step, like here) by computing
```
len(t0.pd_preprocess) + index_that_failed_on_the_forward = 2 + 1= 3
```
When the preprocess part is a `Rollout` or a `Parallel` another logic is applied. For example, let `t1` be 
```
to_tensor = transform(torch.LongTensor(x)
prep = (to_tensor >> batch) / (to_tensor >> batch)
t1 = (prep
    >> transform(torch.nn.Embedding)(10, 8) / identity
    >> transform(torch.nn.Linear)(6, 4) / identity
)
```
This is failing on the step 2 of the transform. For example, after executing `t1.infer_apply(([9, 8, 8], [1]))` and calling the debugger, we get this
```
Compose - "t1":

      │└──────────────────────────────────────────────┐
      │                                               │
      ▼ args                                          ▼ args
   0: [to_tensor >> Batchify(dim=0)] / [to_tensor >> Batchify(dim=0)]
      │                                               │
      ▼ input                                         ▼ args
   1: Embedding(num_embeddings=10, embedding_dim=8) / Identity()                    
      │                                               │
      ▼ input                                         ▼ args
   2: Linear(in_features=6, out_features=4)         / Identity()                      <---- error here
```
The way we compute  where the arrow points in this case (when the whole transform is a `Compose` and `pd_preprocess` is a `Parallel` or a `Rollout`) is by computing
```
1 + index_that_fails_on_the_forward = 1 + 1
```
This is the way we had been doing but there is a lack of generality on this criterium. It is when the transform is a `Compose`, the `preprocess` is a `Parallel` or `Rollout`, and the preprocess and forward are mixed on the same step of the transform. For example, let `t2` be
```
data_pipeline = to_tensor >> batch >> transform(torch.nn.Embedding)(10, 8) >> transform(torch.nn.Linear)(6, 4)
target_pipeline = to_tensor >> batch
t2 = (data_pipeline / target_pipeline) >> transform(torch.nn.CrossEntropyLoss)()
```
Now, if we execute `t2` (which fails on the linear layer) and call the debugger we get 
```
Compose - "t":

      │└────────────────────────┐
      │                         │
      ▼ args                    ▼ args
   0: [data_pipeline: ..>>..] / [to_tensor >> Batchify(dim=0)]
      │
      ▼ (input, target)
   1: CrossEntropyLoss()       <---- error here 
```
Why is it not pointing to 0 like expected? It is because the previous logic made for`t1` does not hold anymore. According to the previous rule, we failing step should be
```
1 + index_that_fails_on_the_forward = 1 + 0
```
since `t2.pd_forward` is
```
Compose:

      │└──────────────────────────────────────────────┐
      │                                               │
      ▼ args                                          ▼ args
   0: [<anonymous Embedding> >> <anonymous Linear>] / Identity()
      │
      ▼ (input, target)
   1: CrossEntropyLoss()
```
Then, the rule applied for `t1` is not correct here. We shouldn't sum 1 to the failing index on the forward, but 0. 

However, it's difficult to hold the cases for `t1`  and `t2` at the same time. We would need to check if the preprocess part and forward are mixed a same item of a `Compose`, which is not straightforward.

eramdiaz:
## Description
Fixes #404 and #340 by not showing anymore the arrow at the highest level of the debugger (where the entire transform we are executing is shown).

For example, let `t` be
```
t = (transform(lambda x: torch.LongTensor(x))
     >> batch
     >> transform(torch.nn.Embedding)(10, 8)
     >> transform(torch.nn.Linear)(6, 4)
)
```
Then, an error will be produced at the linear layer due to mismatching shapes. If we execute and call the debugger, at the first level we will get 
```
Compose - "t":

      │
      ▼ x
   0: lambda x: torch.LongTensor(x)                
      │
      ▼ args
   1: Batchify(dim=0)                              
      │
      ▼ input
   2: Embedding(num_embeddings=10, embedding_dim=8)
      │
      ▼ input
   3: Linear(in_features=6, out_features=4)      
```
as we see with no marker arrow. The deeper levels will already provide the marker arrow.  For example, the next one is
```

Compose:

      │
      ▼ input
   0: Embedding(num_embeddings=10, embedding_dim=8)
      │
      ▼ input
   1: Linear(in_features=6, out_features=4)          <---- error here 
```

### DIFF
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -537,14 +537,6 @@ class Transform:
             break
         return f'{a_tb.filename} in {a_tb.name}\n----> {make_green(a_tb.lineno)}    {a_tb.line}'
 
-    def _pd_get_error_idx(self, stage: str):
-        """Get what element of a :class:`padl.transforms.Transform` is failing if an Exception is
-        produced during an execution.
-
-        Subclasses of :class:`padl.transforms.Transform` need to implement this method.
-        """
-        return NotImplemented
-
     def _pd_trace_error(self, position: int, arg):
         """Add some error description to :obj:`pd_trace`. """
         try:
@@ -830,8 +822,7 @@ class Transform:
                     try:
                         output = forward.pd_call_in_mode(batch, mode)
                     except Exception as err:
-                        self._pd_trace_error(self._pd_get_error_idx('forward'),
-                                             [args[i] for i in ix])
+                        self._pd_trace_error(None, [args[i] for i in ix])
                         raise err
 
                 if use_post or flatten:
@@ -841,8 +832,7 @@ class Transform:
                             output = [post.pd_call_in_mode(x, mode, ignore_grad=True)
                                       for x in output]
                         except Exception as err:
-                            self._pd_trace_error(self._pd_get_error_idx('postprocess'),
-                                                 [args[i] for i in ix])
+                            self._pd_trace_error(None, [args[i] for i in ix])
                             raise err
 
                     for out in output:
@@ -969,13 +959,13 @@ class Transform:
             try:
                 inputs = forward.pd_call_in_mode(inputs, mode='infer')
             except Exception as err:
-                self._pd_trace_error(self._pd_get_error_idx('forward'), in_args)
+                self._pd_trace_error(None, in_args)
                 raise err
         if use_post:
             try:
                 inputs = postprocess.pd_call_in_mode(inputs, mode='infer', ignore_grad=True)
             except Exception as err:
-                self._pd_trace_error(self._pd_get_error_idx('postprocess'), in_args)
+                self._pd_trace_error(None, in_args)
                 raise err
         return self._pd_format_output(inputs)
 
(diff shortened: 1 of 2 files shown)


wuhu:
Why did you decide to remove the top level marker?

eramdiaz:
This is supposed to solve the issue #404 (and #340), which describes a bug on the top level marker in a non edgy case. It is not a direct fix but I don't know how to solve it in a direct way. Let me know if you agree with the fix @wuhu :)

eramdiaz:
This is supposed to solve the issue #404 (and #340), which describes a bug on the top level marker in a non edgy case. It is not a direct fix but I don't know how to solve it in a direct way. Let me know if you agree with the fix @wuhu :)

wuhu:
@eramdiaz I got that, I was trying to understand why you want to remove the marker entirely on the top level, rather than making it work in those cases - Is it not possible to fix it in a direct way or do you think it's too complicated / not worth the effort?

PR #409, opened 2022-02-23
403 variables

### LINKED ISSUE #403: Support for old variables
## 🛰️ Feature

In `aleph` we had variables which could be specified at load time, and which were not saved with the model (because that didn't necessarily always make sense.

At load time we had `load(path, some_variable_to_be_changed=<value>)`. This makes sense for many training scenarios.

wuhu:
Fixes #403

From new docs:

Using `padl.param` one can specify parameters which can then be overridden when loading files.

```python
from padl import transform, param, load

x = param(1, name='x', description='add this much')

@transform
def f(y):
  return x + f

save(f, 'f.padl')
```

When loading a Transform that uses a parameter, one can specify the value of the parameter as
a keyword argument to `padl.load`:

```python
>>> from padl import load
>>> f = load('f.padl', x=1000)
>>> f(1)
1001
```

Params usually have a default value. One can force the user to provide them with
`use_default=False`.


docs/src/advanced/saving.md
```diff
@@ -255,6 +255,35 @@ x = value(x, (mysaver, myloader))
...
+Using {func}`padl.param` one can specify parameters which can then be overridden when loading files.
+
+```python
+from padl import transform, param, load
+
+x = param(1, name='x', description='add this much')
+
+@transform
+def f(y):
+  return x + f
```
jasonkhadka, line 269: x + y

docs/src/advanced/saving.md
```diff
@@ -255,6 +255,35 @@ x = value(x, (mysaver, myloader))
...
+
+save(f, 'f.padl')
+```
+
+When loading a Transform that uses a parameter, one can specify the value of the parameter as
```
jasonkhadka, line 274: new line is not needed here.

padl/dumptools/serialize.py
```diff
@@ -133,3 +133,41 @@ def value(val, serializer=None):
     sourceget.put_into_cache(caller_frameinfo.filename, sourceget.original(source),
                              _serialize(val, serializer=serializer), *locs)
     return val
+
+
+def param(val, name, description=None, use_default=True):
+    """Helper function that marks things in the code are parameters.
```
jasonkhadka, line 139: `that` is missing

jasonkhadka:
MIssing tests, except that looks good.

jasonkhadka:
I find this behaviour confusing. 


```
factor = param(2, name='x', description='multiply this much')


@padl.transform
def plus(*a):
    return sum(list(a))

@padl.transform
class multiply:
    def __init__(self, factor):
        self.x = factor
    
    def __call__(self, arg):
        return arg*self.x
    
@padl.transform
def plus_factor(x):
    return x+factor
```

```
m1 = multiply(factor)
m2 = multiply(factor)
t = plus_factor >> m1 + m2 >> plus

save(t, 't.padl', force_overwrite=True)
```

```

t_load = load('t.padl', factor=20)
t_load(1)
12

t_load

Compose:

      │
      ▼ x
   0: plus_factor       
      └────────────────────┐
      │                    │
      ▼ arg                ▼ arg
   1: multiply(factor=2) + multiply(factor=2)
      │
      ▼ a
   2: plus  
```

```

t_load2 = load('t.padl', x=20)
t_load2(1) 
840

t_load2
Compose:

      │
      ▼ x
   0: plus_factor        
      └─────────────────────┐
      │                     │
      ▼ arg                 ▼ arg
   1: multiply(factor=20) + multiply(factor=20)
      │
      ▼ a
   2: plus      
```

In the printout, it shows as `factor` (the variable name) but while loading you have to load it with `param name`.

wuhu:
@jasonkhadka thanks for the review.

On your last comment:

The printout shows the keyword argument name of `multiply`'s `__init__`, not the variable name. We cannot make the param use that because we want to be able to use a param in more than one place:

```python
x = param(123, 'x')

@transform
class A:
    def __init__(self, kwarg_of_A):
        [...]

@transform
class add:
    def __init__(self, kwarg_of_B):
        [...]

t = (
    A(x) >> B(x)
)

save(b, ...)
```

Another option would be to use the variable name instead of having to give a name (`x = param(123)` would have x), but that involves some trickery with the potential of new bugs. Also it would mean that one could only use param in the context of a variable assignment.

Do you have another suggestion what we could do about this?

docs/src/advanced/saving.md
wuhu: :+1:

wuhu:
Added tests.

jasonkhadka:
Thinking about it, I think this should fail:
```
t_load = load('t.padl', factor=20)
```
as `param` is defined by
```
factor = param(2, name='x', description='multiply this much')
```
Currently, the behaviour is hiding the problem with `load(...,factor=20)`. I am not setting the correct value to the intended parameter, but i am not becoming aware of that.  

If we get error like:
```
t_load = load('t.padl', factor=20)
ERROR: Parameter 'factor' does not exist. t.padl contains parameters: (`x`, default_value=2)
```
I think this should be enough to aware of people about the difference between the variable name and parameter name.  

> @jasonkhadka thanks for the review.
> 
> On your last comment:
> 
> The printout shows the keyword argument name of `multiply`'s `__init__`, not the variable name. We cannot make the param use that because we want to be able to use a param in more than one place:
> 
> ```python
> x = param(123, 'x')
> 
> @transform
> class A:
>     def __init__(self, kwarg_of_A):
>         [...]
> 
> @transform
> class add:
>     def __init__(self, kwarg_of_B):
>         [...]
> 
> t = (
>     A(x) >> B(x)
> )
> 
> save(b, ...)
> ```
> 
> Another option would be to use the variable name instead of having to give a name (`x = param(123)` would have x), but that involves some trickery with the potential of new bugs. Also it would mean that one could only use param in the context of a variable assignment.
> 
> Do you have another suggestion what we could do about this?

wuhu:
> If we get error like:
> 
> ```
> t_load = load('t.padl', factor=20)
> ERROR: Parameter 'factor' does not exist. t.padl contains parameters: (`x`, default_value=2)
> ```
> 
> I think this should be enough to aware of people about the difference between the variable name and parameter name.
> 

good idea, I'll add that.

PR #425, opened 2022-03-16
Generate real requirements.txt instead of versions.txt

wuhu:
- creates a requirements.txt instead of a versions txt
- this uses importlib_metadata.packages_distributions to find the correct name and version
- if a package cannot be found (for example if it's not installed), an exception is raised, this can be suppressed using `save(..., strict_requirements=False)`


sjrl (approved):
Looks good! Could you add a test example that throws the `RequirementNotFound` exception?

PR #428, opened 2022-03-18
Simplify get segment

wuhu:
This PR simplifies `inspector.get_segment_from_frame`, function responsible to extract code segments from a frame object.
Previously, for run-time reasons, this would first try to get the block of code surrounding the statement before extracting it. This proved to be the cause of many bugs, therefore I removed it.
This makes loading code a bit slower. To mitigate, I added a cached version of `ast.parse`. Tests run a bit (factor of ~1.5) slower, but in practice this shouldn't be a problem.


padl/dumptools/inspector.py
```diff
@@ -352,70 +276,51 @@ def get_segment_from_frame(caller_frame: types.FrameType, segment_type, return_l
...
     # be present in instrs)
     target_instrs = [x for x in target_instrs if x.opname != 'EXTENDED_ARG']
 
-    segment = None
-    locs = None
-    found = False
-
-    for segment, locs in sorted(candidate_segments, key=lambda x: -len(x[0])):
-
+    def check_segment(segment):
```
sjrl, line 294: Could we make this a function outside of the scope of `get_segment_from_frame`? I think it will make it easier to read as a new developer.

sjrl, line 294: Could we make this a function outside of the scope of `get_segment_from_frame`? I think it will make it easier to read as a new developer.

padl/dumptools/ast_utils.py
```diff
@@ -19,6 +20,13 @@ class Position:
     end_col_offset: int
 
 
+def cached_parse(source):
```
sjrl, line 23: I realize it is a very simple function, but I would appreciate a short docstring explaining what it is for.

padl/dumptools/inspector.py
wuhu: Not sure I agree, can you explain why you think it would make it easier to read? I think there is some benefit in having it close to the one place where it's being used.

sjrl: I see. Yes I agree that having it closer to where it's used can be helpful. For me it becomes easier to see the dependencies of a function if all external variables used inside the function are passed to it through the function call. Similar to how you built `instructions_are_the_same`.

sjrl: okay sounds good

padl/dumptools/inspector.py
```diff
@@ -323,6 +218,35 @@ def caller_frame() -> types.FrameType:
     return outer_caller_frameinfo(calling_module_name).frame
 
 
+def instructions_are_the_same(instr, target_instr, frame=None):
+    """Check if the disassebled instructions *instr* and *target_instr* are (basically) the same.
```
jasonkhadka, line 222: `disassebled`? Is it `disassembled`?

padl/dumptools/inspector.py
```diff
@@ -323,6 +218,35 @@ def caller_frame() -> types.FrameType:
...
+
+    Mainly this checks if the *opname* and *argval* of the two instructions are the same,
+    additionally accounting for some subleties involving code object - argvals and LOAD-operators.
```
jasonkhadka, line 225: subtleties?

padl/dumptools/inspector.py
```diff
@@ -323,6 +218,35 @@ def caller_frame() -> types.FrameType:
     return outer_caller_frameinfo(calling_module_name).frame
 
 
+def instructions_are_the_same(instr, target_instr, frame=None):
```
jasonkhadka, line 221: Would be great to have docstrings for the parameters.

padl/dumptools/inspector.py
```diff
@@ -323,6 +218,35 @@ def caller_frame() -> types.FrameType:
...
+    if instr.opname == target_instr.opname:
+        return True
+
+    # load operations can have many different name, accept if both are load ops regardless
+    # of which kind exactly
+    load_ops = ('LOAD_NAME', 'LOAD_FAST', 'LOAD_GLOBAL', 'LOAD_CONST', 'LOAD_DEREF')
+    return instr.opname in load_ops and target_instr.opname in load_ops
+
+
 def get_segment_from_frame(caller_frame: types.FrameType, segment_type, return_locs=False) -> str:
```
jasonkhadka, line 250: I think there should be `assertion` for `segment_type`. Cause if there is typo in `segment_type` or something else is passed, things will fail pretty far down in the code.

jasonkhadka (changes requested):
Some small comments.

padl/dumptools/inspector.py
wuhu: yes

padl/dumptools/inspector.py
wuhu: :+1:

padl/dumptools/inspector.py
wuhu: makes sense

wuhu:
@jasonkhadka thanks for the review, I've changed accordingly.

jasonkhadka (approved):

tests/unittests/dumptools/test_inspector.py
```diff
@@ -30,145 +30,3 @@ def test_nested(self):
     def test_double_nested(self):
         with pytest.raises(AssertionError):
             double_nested_builder()
-
-
-testcode_0 = '''
```
sjrl, line 35: @wuhu Would it still be good to have these examples to test exotic formatting?

sjrl (approved):

PR #430, opened 2022-03-18
Random cleanup

wuhu:
Some random cleanup + linting.


blythed (approved):
Looks ok to me. I'm guessing this is the caching we were talking about.

Issue #236, opened 2021-11-12
force overwrite in saving


eramdiaz:
## Description

Force overwrite should not delete the previous folder if the saving task was not completed succesfully.

PR #431, opened 2022-03-18
Force overwrite

wuhu:
#236


padl/transforms.py
```diff
@@ -405,12 +405,16 @@ def pd_save(self, path: Union[Path, str], force_overwrite: bool = False,
...
+        if path.exists() and list(path.glob('*')):
             if not force_overwrite:
                 raise FileExistsError(f'{path} exists, call with *force_overwrite* to overwrite.')
-            rmtree(path)
 
-        path.mkdir()
+            with TemporaryDirectory('.padl') as dirname:
+                self.pd_save(dirname, False)
+                rmtree(path)
+                copytree(dirname, path)
```
sjrl, line 415: I think it could be a good idea to use a `move` method instead to avoid the situation where two copies of the weights exist at one time. This potentially could be a problem when working with large models. What do you think?

sjrl:
Hmm I'm not entirely sure how the master merge I performed is causing the `RequirementNotFound` errors. @wuhu Do you know why the `padl` package is not being found for these tests?

padl/transforms.py
sjrl: Never mind I don't think this will be a practical problem people have and I don't think it's worth changing.

jasonkhadka (approved):

Issue #427, opened 2022-03-18
Assignment breaks the saving 


jasonkhadka:
## 🐞 Bug

Assignment breaks the saving

## Works BUT
* `pd_to` is not written in the `transform.py` file
* `visit_Attribute` is not triggered.
```
DEVICE = 'cpu'


@transform
def times_two(x):
    return x * 2

times_two.pd_to(DEVICE)

save(times_two, 'm.padl')
m = load('m.padl')
```
content of `transform.py`

```
from padl import transform


@transform
def times_two(x):
    return x * 2


_pd_main = times_two
```

## Not working code
* triggers `visit_Attribute` 6 times.
* creation of `times_two` is not written on the file.
```
DEVICE = 'cpu'


@transform
def times_two(x):
    return x * 2

times_two = times_two.pd_to(DEVICE)

save(times_two, 'm.padl')
m = load('m.padl')
```

content of `transform.py`
```
DEVICE = 'cpu'
times_two = times_two.pd_to(DEVICE)
_pd_main = times_two
```

## Works BUT
* forgets about `device` assigment.
* forgets about `new_name`
* does not trigger `visit_Attribute`

```
import padl
from padl import transform, save, load

DEVICE = 'cpu'

@transform
def times_two(x):
    return x * 2

new_name = times_two.pd_to(DEVICE)

save(new_name, 'm.padl', force_overwrite=True)
m2 = load('m.padl')
```
This works. but forgets about `new_name` and `new_name` assigment.
Below is the code written on `transform.py`.
```
from padl import transform


@transform
def times_two(x):
    return x * 2


_pd_main = times_two
```

###

sjrl:
- Should the saved source code that is written to `transform.py` be checked at save time to make sure that the save worked? We could potentially allow users to pass a test input to make sure that the code works as expected at loading time.

wuhu:
@sjrl I think the automatic check is a good idea, at least for now so that if something isn't saved properly it doesn't fail silently.
Just saving the layers could also be useful.

@jasonkhadka the two "Works BUT" examples are expected:
(1) Saving does not dump attribute calls if they are not part of an assignment (see https://lf1-io.github.io/padl/latest/advanced/saving.html#mutated-objects)
(2) function transforms always go by the name of the function, not by the name of a variable they're assigned to, so when you dump them, the saver tracks down the name of the function, not the name of the variable - so it's not strictly a bug, still we could think about whether this is the most intuitive behavior and perhaps change it

[...]

PR #432, opened 2022-03-21
fix by checking for increment

jasonkhadka:
## Description

Fixes [#427
](https://github.com/lf1-io/padl/issues/427)

Checks for increasing the `scopedname.n` is implemented.


jasonkhadka:
Check the continued discussion on the issue: https://github.com/lf1-io/padl/issues/427

wuhu:
Thanks for this!

Could you explain why your changes fix the problem?

Also I found that it doesn't seem to work in combination with recursive function transforms, e.g.:

```python
from padl import *
@transform
def f(x):
    return f(x-1)


f = f.pd_to('cuda')
print(f._pd_dumps())
```

raises an exception.

jasonkhadka:
very strange!!
Cause this works:
```
from padl import *
@transform
def f(x):
    return 1 + f(x-1)


#f = f.pd_to('cuda')
print(f._pd_dumps())
```

I will have a look tomorrow, and also write the explanation.

> ```python
> from padl import *
> @transform
> def f(x):
>     return f(x-1)
> 
> 
> f = f.pd_to('cuda')
> print(f._pd_dumps())
> ```

padl/dumptools/var2mod.py
```diff
@@ -596,8 +597,30 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
     return globals_
 
 
+def _check_and_make_increment(var: ScopedName, scope: Scope, scoped_name: ScopedName):
```
sjrl, line 600: I have a question. Do we need both `scope` and `scoped_name` here? I thought `scoped_name` also contains the `scope` of the target.

sjrl: Ahh I see further down how these can differ. Could you add to the `scope` doc string why this could differ from `scoped_name`?

padl/transforms.py
```diff
@@ -526,17 +526,15 @@ def f(x):
 
             # see if the object itself knows how to generate its codegraph
             try:
-                if next_name.scope.is_global():
-                    next_obj = globals_dict[next_name.name]
-                else:
-                    next_obj = all_vars_dict[next_name.name]
+                next_obj = all_vars_dict[next_name.name]
```
sjrl, line 529: @wuhu This is a change @jasonkhadka and I discussed since it looks like `all_vars_dict` already contains the `globals_dict`. Was there any other reason to have this `if` statement?

padl/dumptools/var2mod.py
```diff
@@ -596,8 +597,30 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
...
+        new_scoped_name = ScopedName(var.name, scope, var.n + scoped_name.n + 1)
+        find_in_scope(new_scoped_name)
+    except NameNotFound:
+        new_scoped_name = ScopedName(var.name, scope, var.n + scoped_name.n)
+    return new_scoped_name
+
+
 def increment_same_name_var(variables: List[ScopedName], scoped_name: ScopedName):
-    """Go through *variables* and increment the the counter for those with the same name as
+    """Go through *variables* and increment the counter for those with the same name as
```
sjrl, line 623: Could you add to the end of the docstring `... with the same name as *scoped_name* by incrementing *scoped_name.n*` to make it more clear what we are doing?

padl/transforms.py
sjrl: Okay thanks for the info! @jasonkhadka Could we add the example provided by @wuhu to the unit tests to test this scenario? Also I think it would be helpful to add a comment in the code explaining this situation or at least that `all_vars_dict` relies on `nonlocals_dict` overwriting the keys in `globals_dict`.

padl/dumptools/var2mod.py
sjrl: Never mind this doesn't make sense.

padl/dumptools/var2mod.py
```diff
@@ -596,8 +597,30 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
     return globals_
 
 
+def _check_and_make_increment(var: ScopedName, scope: Scope, scoped_name: ScopedName):
+    """Check and make increment if the ScopedName after increment exists.
```
sjrl, line 601: I would reword this to `Increment ScopedName and check if it exists.`

padl/dumptools/var2mod.py
```diff
@@ -182,6 +189,7 @@ def visit_Attribute(self, node):
 
         scope = getattr(node, '_scope', None)
         name = ScopedName('.'.join(path), scope, 0)
+        # name = ScopedName('.'.join(path), scope, 0)
```
sjrl, line 192: I think you can remove this extra comment

padl/dumptools/var2mod.py
```diff
@@ -611,11 +619,11 @@ def _check_and_make_increment(var: ScopedName, scope: Scope, scoped_name: Scoped
     :return:
         ScopedName after increment
     """
+    new_scoped_name = ScopedName(scoped_name.name, scope, var.n + scoped_name.n + 1)
     try:
-        new_scoped_name = ScopedName(var.name, scope, var.n + scoped_name.n + 1)
         find_in_scope(new_scoped_name)
     except NameNotFound:
-        new_scoped_name = ScopedName(var.name, scope, var.n + scoped_name.n)
+        new_scoped_name = ScopedName(scoped_name.name, scope, var.n + scoped_name.n)
```
sjrl, line 626: Why are we using `scoped_name.name` here instead of `var.name`?

padl/transforms.py
jasonkhadka: Then the question becomes if there are cases when the scope is not global for an object but it is in `globals_dict`. I think this can be the case when we are looking at some internal scope, where variables from the outer scope are used, but not necessarily global. 
Is this the case @wuhu? 
If not, then we can skip defining `all_vars_dict`, and use `non_local_dict` in `else` clause.

padl/dumptools/var2mod.py
jasonkhadka: I updated this using `var.name`.  `var.name` makes more sense.

padl/transforms.py
jasonkhadka: Thanks for the clarifaction. I reverted this change.

jasonkhadka:
I have removed doctest on `increament_same_name_var`. The example is still there.

padl/dumptools/var2mod.py
jasonkhadka: i moved scope to _check_and_make_increment, and added some comment to explain the things there.

sjrl:
@jasonkhadka What was wrong with the doctest?

jasonkhadka:
`find_in_source` did not work with this doctest code. It is about finding the right scope. 
No fundamental issue here, just think we do not need doctest here.

> @jasonkhadka What was wrong with the doctest?

sjrl:
@jasonkhadka Could you add the examples @wuhu showed here https://github.com/lf1-io/padl/pull/432#discussion_r832318191 as unit tests? I think it would be good to test the behavior of the `if else` statement in question.

sjrl (approved):
Looks good!

jasonkhadka:
> 

yes sure!

jasonkhadka:
> @jasonkhadka Could you add the examples @wuhu showed here https://github.com/lf1-io/padl/pull/432#discussion_r832318191 as unit tests? I think it would be good to test the behavior of the if else statement in question.

padl/dumptools/var2mod.py
```diff
@@ -597,28 +605,52 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
...
+def _check_and_make_increment(value: ScopedName, targets: Set):
+    """Apply Increment on *value* ScopedName if necessary.
+
+    Increment means the increment of *value.n*.
+    *value.n* signifies which one of the usages of a *name* in the code does this
+     *scoped_name* points to.
+    0 being the most recent, and increase integer representing the older usages.
+
+    :param value: Target ScopedName for increment
+    :param targets: Set of ScopedName for increment
```
sjrl, line 617: I would consider rephrasing the docstrings to something like 
```python
:param value: ScopedName to be incremented (if necessary)
:param targets: Set of ScopedName's to compare to *value*
```
Right now it's a little confusing that `value` is described as the `Target ScopedName` when we also have a parameter called `targets`.

wuhu (changes requested):
This is going in the right direction, main problem I think is what I wrote to you about on slack (and here, line 621).

padl/dumptools/var2mod.py
jasonkhadka: thanks updated it.

padl/dumptools/var2mod.py
```diff
@@ -62,13 +62,20 @@ def get_source_segments(self, source):
         ]
 
 
-def _join_attr(node):
-    if not isinstance(node, (ast.Attribute, ast.Name)):
+def _join_attr(ast_node):
+    """Get list of base object name and attribute names
+
+    'f.pd_to' -> ['f', 'pd_to']
```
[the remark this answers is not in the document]
jasonkhadka: sure.

padl/dumptools/var2mod.py
```diff
@@ -597,28 +605,52 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
...
+    *value.n* signifies which one of the usages of a *name* in the code does this
+     *scoped_name* points to.
+    0 being the most recent, and increase integer representing the older usages.
+
+    :param value: Target ScopedName for increment
+    :param targets: Set of ScopedName for increment
+    :return:
+        ScopedName after increment
+    """
+    split_value_name = value.name.rsplit('.', 1)[0]
```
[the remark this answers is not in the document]
jasonkhadka: I was also thinking about this problem.
I am continuing this discussion on the issue: https://github.com/lf1-io/padl/issues/427

padl/dumptools/symfinder.py
```diff
@@ -668,14 +681,65 @@ def replace_star_imports(tree: ast.Module):
...
+    for variant_name, variant_n in scoped_name.variants:
+        if remove_dot and '.' in variant_name:
+            variant_name = variant_name.rsplit('.', 1)[0]
+        return_scoped_name.add_variant(variant_name, add_n + variant_n)
+    return_scoped_name.variants.pop(0)
+    return_scoped_name.name, return_scoped_name.n = return_scoped_name.variants[0]
+    return return_scoped_name
+
+
+def find_scopedname_in_source(scoped_name: ScopedName, source, tree=None) -> Tuple[str, ast.AST]:
```
sjrl, line 703: Update return type to show that `Tuple` returns three items.

padl/dumptools/symfinder.py
```diff
@@ -668,14 +681,65 @@ def replace_star_imports(tree: ast.Module):
...
+    """Find the piece of code that assigned a value to the variable with name *var_name* in the
+    source string *source*.
+
+    :param scoped_name: ScopedName to look for.
+    :param tree: AST.Module to look into for scoped_name.
+    :param source: Source code to search.
+    :returns: Tuple with source code segment and corresponding AST node.
```
sjrl, line 710: Update docstring to reflect returning three things.

padl/dumptools/symfinder.py
```diff
@@ -736,6 +805,24 @@ def _find_branch(tree, lineno, source):
...
+        try:
+            source, node, name = find_scopedname_in_source(scoped_name, cell)
+        except (NameNotFound, SyntaxError):
+            continue
+        break
+    if source is None:
+        raise NameNotFound(f'{",".join([str((var, n)) for var, n in scoped_name.variants])} not found.')
+    return source, node, name
+
+
```
sjrl, line 825: @jasonkhadka Do we need both `find_in_ipython` and `find_scopedname_in_ipython`?

padl/dumptools/symfinder.py
```diff
@@ -607,39 +608,51 @@ def f(x):
...
+    :return: Tuple as ((source, node), scope, name), where
+        * source: String representation of piece of code.
+        * node: Ast node for the code.
+        * scope: Scope of the code.
+        * name: Name of variable (str).
+
     """
     scope = scoped_name.scope
-    i = scoped_name.n
+    n = scoped_name.n
```
sjrl, line 642: I think you can remove this line. It looks like `n` is not used in this function.

padl/dumptools/symfinder.py
```diff
@@ -759,6 +846,29 @@ def find_in_ipython(var_name: str, i: int = 0) -> Tuple[str, ast.AST]:
...
 
 
+def find_scopedname(scoped_name: ScopedName) -> Tuple[str, ast.AST]:
+    """Find the piece of code that assigned a value to the variable with name *var_name* in the
+    module *module*.
+
+    If *module* is not specified, this uses `__main__`. In that case, the ipython history will
+    be searched as well.
+
+    :param var_name: Name of the variable to look for.
```
sjrl, line 856: Remove doc string for `var_name`

sjrl: * Same for `module` and `i`.
* Add doc string for `scoped_name`

padl/dumptools/symfinder.py
```diff
@@ -759,6 +846,29 @@ def find_in_ipython(var_name: str, i: int = 0) -> Tuple[str, ast.AST]:
     return source, node, name
 
 
+def find_scopedname(scoped_name: ScopedName) -> Tuple[str, ast.AST]:
+    """Find the piece of code that assigned a value to the variable with name *var_name* in the
+    module *module*.
```
sjrl, line 851: Update doc string to reflect the different input variable

padl/dumptools/symfinder.py
jasonkhadka: removed!

padl/dumptools/symfinder.py
jasonkhadka: I am not sure yet, 
I don't see another use at the moment for `find_in_ipython`, but it is useful function so could be used or used for debugging.

sjrl: I guess in general it looks like there are a few functions that look almost duplicated like `find_scopedname_in_module` and `find_in_module`, and `find_scopedname_in_source` and `find_in_source`. Is it possible to just use the new `find_scopedname_` versions?

jasonkhadka: Yes, I updated the finders to use `find_scopedname*` instead of `find_*`, cause it contains more information. I am not sure about throwing out `find_*`s yet. They can be used in debugging at the least and perhaps will be more useful for other new features. Also, it is not a whole lot of code.

padl/dumptools/var2mod.py
```diff
@@ -597,31 +607,74 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
...
+     *scoped_name* points to.
+    0 being the most recent, and increase integer representing the older usages.
+
+    :param value: ScopedName to be incremented (if necessary).
+    :param targets: Set of ScopedName's to compare to *value*.
+    :return: ScopedName after increment.
+    """
+    split_value_name = value.name.rsplit('.', 1)[0]
+    for target in targets:
+        split_target_name = target.name.rsplit('.', 1)[0]
```
wuhu, line 624: I think this should cover all `.`s, not just the rightmost one.

padl/dumptools/var2mod.py
```diff
@@ -597,31 +607,74 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
...
-    >>> isinstance(out, set)
+    >>> import sys
+    >>> a = b = 2
+    >>> b = a + b
+    >>> out = increment_same_name_var({ScopedName('a', None, 0), ScopedName('b', None, 0)},
+    ...                 ScopedName('b', Scope.toplevel(sys.modules[__name__]), 0))  # doctest: +SKIP
+    >>> isinstance(out, set)  # doctest: +SKIP
     True
-    >>> {(x.name, x.n) for x in out} == {('a', 1), ('b', 4)}
+    >>> {(x.name, x.n) for x in out} == {('a', 0), ('b', 1)}  # doctest: +SKIP
```
wuhu, line 664: Why still skipping, does it not work?

padl/dumptools/symfinder.py
```diff
@@ -668,14 +680,65 @@ def replace_star_imports(tree: ast.Module):
                 node.names = [ast.alias(name=name, asname=None) for name in names]
 
 
+def update_scopedname(scoped_name: ScopedName, scope: Scope, add_n: int = 0, remove_dot: bool = False):
```
wuhu: You only ever use this with `add_n != 0 and remove_dot == False` or `remove_dot == True and add_n == 0`.
It would be more readable if you had two separate functions `remove_dot` and `add_n` (or something like that). (If `remove_dot` is even needed, see comment in `find_scopedname_in_source`).

padl/dumptools/symfinder.py
```diff
@@ -607,39 +608,50 @@ def f(x):
...
+    def add_variant(self, name, n):
+        self.variants.append((name, n))
+
     def __hash__(self):
         return hash((self.name, self.scope, self.n))
 
     def __eq__(self, other):
         return (self.name, self.scope, self.n) == (other.name, other.scope, other.n)
 
+    def copy(self):
```
wuhu, line 623: Is this ever used?

padl/dumptools/symfinder.py
```diff
@@ -607,39 +608,50 @@ def f(x):
     scope: Scope
     n: int = 0
```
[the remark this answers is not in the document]
wuhu: Also please add some explanation about the `variants` in the docstring.

padl/dumptools/var2mod.py
```diff
@@ -588,6 +597,7 @@ def find_globals(node: ast.AST, filter_builtins: bool = True) -> Set[Tuple[str,
 
     :param node: AST node to search in.
     :param filter_builtins: If *True*, filter out builtins.
+    :return: Set of global ScopedNames
```
wuhu, line 600: .

padl/wrap.py
```diff
@@ -19,6 +19,7 @@
 )
 
 
+
```
wuhu, line 22: Remove this extra line.

wuhu (changes requested):

padl/dumptools/var2mod.py
jasonkhadka: yes, still does not work. Something to do with Scope.

padl/dumptools/symfinder.py
jasonkhadka: it is used with refactor.

padl/dumptools/symfinder.py
jasonkhadka: I have moved this to the methods of ScopedName.

padl/dumptools/symfinder.py
```diff
@@ -668,14 +680,65 @@ def replace_star_imports(tree: ast.Module):
...
+    :param add_n: *n* to add to new ScopedName.
+    :param remove_dot: Boolean to remove dot from ScopedName or not.
+    :return: New ScopedName with updated name and *n*.
+    """
+    return_scoped_name = ScopedName(None, scope, None)
+    for variant_name, variant_n in scoped_name.variants:
+        if remove_dot and '.' in variant_name:
+            variant_name = variant_name.rsplit('.', 1)[0]
+        return_scoped_name.add_variant(variant_name, add_n + variant_n)
+    return_scoped_name.variants.pop(0)
```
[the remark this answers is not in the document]
jasonkhadka: i refactored ScopedName to make it clearer.

padl/dumptools/symfinder.py
```diff
@@ -668,14 +680,65 @@ def replace_star_imports(tree: ast.Module):
...
+    for statement in tree.body[::-1]:
+        for var_name, var_n in scoped_name.variants:
+            for finder_cls in finder_clss:
+                finder = finder_cls(source, var_name)
+                finder.visit(statement)
+                if finder.found_something():
+                    if n_dict[var_name] == 0:
+                        return finder.deparse(), finder.node(), var_name
+                    n_dict[var_name] -= 1
+    if any(['.' in var_name for var_name, _ in scoped_name.variants]):
```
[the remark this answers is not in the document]
jasonkhadka: Refactored to fix this. 
Now, dump is 
```
from padl import transform


class X:
    def __init__(self, x):
        self.random = x


numpy = X(2)
_pd_main = transform(lambda x: numpy.random)
```

padl/dumptools/symfinder.py
jasonkhadka: I refactored this.

padl/dumptools/symfinder.py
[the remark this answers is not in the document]
jasonkhadka: Right now I have kept dataclass for the convenience of the printing. I don't see any harm keeping it or big advantage removing it.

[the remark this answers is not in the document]
wuhu: Yeah, the point was that now that the class has so much extra stuff there's also not that much of a benefit any more, and in my view if there's not much of an advantage it's better to just have an ordinary class (because people are more familiar with it).

wuhu:
@jasonkhadka did you the "5 hidden conversations"?

jasonkhadka:
@wuhu hadn't seen them, but the refactor takes care of all the points.

> @jasonkhadka did you the "5 hidden conversations"?

padl/transforms.py
```diff
@@ -1692,19 +1710,21 @@ def _pd_build_codegraph(self, graph=None, name=None):
...
+
+        todo = self._pd_codegraph_add_startnodes(graph, name)
 
         # iterate over sub-transforms and update the codegraph with their codegraphs
         for transform in self.transforms:
             varname = transform.pd_varname(self._pd_call_info.scope)
             # pylint: disable=protected-access
             transform._pd_build_codegraph(graph, varname)
+
+        #self._pd_codegraph_find_dependencies(graph, todo)
```
sjrl, line 1727: Consider removing commented out code

PR #433, opened 2022-03-22
Docs update

sjrl:
## Description

Update the docs to tell users that PADL is not expected to work in the base Python Interpreter environment. This has to do with frames not containing the source code needed by the `inspect` module.  This is explained in a bit more detail here: https://stackoverflow.com/questions/63347543/python-inspect-stack-returns-frameinfo-objects-with-missing-code-context-a

@wuhu @jasonkhadka Please let me know if we should reword the warning or if there is a more appropriate place to put it.


README.md
```diff
@@ -38,6 +38,12 @@ PADL currently supports python **3.7**, **3.8** and **3.9**.
 Python version >= **3.8** is preferred because creating and loading transforms (**not** execution) 
 can be slower in **3.7**.
 
+WARNING: PADL transforms do not work in the base Python Interpreter environment because we rely on
+the inspect module to find source code (used when saving PADL transforms). Unfortunately, the 
+source code typed at this interactive prompt is discarded as soon as it is parsed. 
+Therefore, we recommend using the iPython interpreter or Jupyter Notebooks for interactive sessions 
+because they do not have this problem.
```
jasonkhadka, line 45: iIthink we can remove 'because they do not have this problem. ' from 
`Therefore, we recommend using the IPython interpreter or Jupyter Notebooks for interactive sessions because they do not have this problem. ` 
and make it 
`Therefore, we recommend using the IPython interpreter or Jupyter Notebooks for interactive sessions.`
Also, 
`iPython` to `IPython`

jasonkhadka:
We can bypass this by avoiding inspection when calling with a normal python console right? We could avoid `call_info` and other `inspections` if it is not possible. Saving will not work there, but anyway working in the normal console is only useful for quick debugging purposes. 
Again a very rare use-case, that we should discuss if we want to support.

sjrl:
> Again a very rare use-case, that we should discuss if we want to support.

I've made an issue for this https://github.com/lf1-io/padl/issues/434

jasonkhadka (approved):
Looks good!

PR #440, opened 2022-03-28
[WIP] Issue 434

### LINKED ISSUE #434: Make PADL work in Base Python Interpreter Environment
## Description

Make PADL to work in environments that do not save executed source code (e.g. in the base Python Interpreter Environment). 
Currently PADL will throw a `NameNotFound` error if the source code of the transform is not findable by the `inspect` module. 
```python
from padl import transform

@transform
def f(x):
    return x + 1
```
gives
```python
```

## Solution

It would be nice if transforms could still be made and executed, but warn the user that saving will not work in this environment. 
We could do this by avoiding inspection when the source code is not available (e.g. returns as `None`) and print a warning to the user that saving will not work. Most likely a rare use-case, but it could be worth supporting. 

## Alternatives

If we do not want to support this we could supplement the `NameNotFound` error to suggest using a different interpreter like IPython if the source code is returned as `None` by the `inspect` module.

sjrl:
## Description

Fixes #434 by adding an error message if the user is in an environment that does not support PADL saving. 

- [ ] Check that this works as expected in Google Colab
- [ ] Check that this works as expected in Jupyter Notebook

### DIFF
--- a/padl/__init__.py
+++ b/padl/__init__.py
@@ -22,14 +22,14 @@ __all__ = ['value', 'transform', 'Batchify', 'Unbatchify', 'Identity', 'group', 
 try:
     import inspect
     from padl.dumptools.sourceget import get_source
-    _ = get_source(inspect.stack()[0].filename)
+    _ = get_source(inspect.stack()[-1].filename)
     # We do not want these to be padl level imports so we remove them
     del inspect
     del get_source
 except Exception as e:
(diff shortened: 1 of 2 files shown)


sjrl:
I think some more thought needs to be put into this. It doesn't look like it is trivial to grab the frame correctly takes input from the interpreter environment. So we can't just use  `_ = get_source(inspect.stack()[-1].filename)`

PR #445, opened 2022-03-29
Add an keyword argument ignore_post on the train_apply method.

### LINKED ISSUE #28: Method to apply preprocess+forward without postprocess
The `trans` part of an `lf` Transform will be now called `preprocess`. We consider this a convenient change since it is clear and gives an inmediate idea of its task. However, we have to give another name to the `trans` + `forward` part. Ideas?

eramdiaz:
## Description
Fixes #28 by adding an keyword argument ignore_post on the train_apply method.

### DIFF
--- a/padl/transforms.py
+++ b/padl/transforms.py
@@ -1017,7 +1017,8 @@ class Transform:
                 raise err
         return self._pd_format_output(inputs)
 
-    def eval_apply(self, inputs: Iterable, flatten: bool = False, **kwargs):
+    def eval_apply(self, inputs: Iterable, flatten: bool = False,
+                   use_post: bool = True, **kwargs):
         """Call transform within the eval context.
 
         This will use multiprocessing for the preprocessing part via `DataLoader` and turn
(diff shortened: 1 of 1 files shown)


padl/transforms.py
```diff
@@ -827,13 +827,14 @@ def _pd_format_output(self, x):
         return x
 
     def _pd_itercall(self, args, mode: Mode, loader_kwargs: Optional[dict] = None,
-                     flatten: bool = False) -> Iterator:
+                     flatten: bool = False, ignore_post:bool = False) -> Iterator:
```
blythed: Missing a space after the `ignore_post`

blythed (commented):
Nice thanks!

blythed (approved):

padl/transforms.py
```diff
@@ -1031,7 +1033,8 @@ def eval_apply(self, inputs: Iterable, flatten: bool = False, **kwargs):
         return self._pd_itercall(inputs, 'eval', loader_kwargs=kwargs,
```
sjrl, line 1033: I would add `use_post` as an option to `eval_apply` as well. It's possible users may want to calculate embedding vectors in bulk instead of returning output of the post processing.

wuhu (approved):
Looks good!

sjrl (approved):

PR #451, opened 2022-04-08
redo params

wuhu:


blythed:
I'm testing this with this code:

```python
from padl import params, transform

p = params('params', A=2, B=2)

def f(A=5, B=6):
    return (
        transform(lambda x: x + A)
        >> transform(lambda x: x + B)
    )

t = f(**p)

t.pd_save('test.padl', force_overwrite=True)
```

It works in `IPython` but not in `python` (when executing the module).

wuhu:
@blythed this should be fixed.

jasonkhadka:
```
from padl import *

@transform
class B:
    def __init__(self, x):
        ...
b = 1
b = B(b)
print(b._pd_dumps())
```

is still failing. 

```
from padl import transform


b = 1


@transform
class B:
    def __init__(self, x):
        ...


b = B(b)
_pd_main = B(b)
```

padl/dumptools/var2mod.py
```diff
@@ -372,150 +379,244 @@ def visit_Import(self, node):
...
+            sub.visit(child)
+            self.res += sub.res
+
+    def visit_Module(self, node):
+        for child in ast.iter_child_nodes(node):
+            self.visit(child)
+
+    def visit_Attribute(self, node):
+        name = ast_utils.get_source_segment(self.source, node)
+        print(name)
```
jasonkhadka, line 478: print statement.

